PR #1, opened 2018-11-03
create chapter directories

lomeli-noe:


lomeli-noe:
yeah, I agree.  Can you delete it?

PR #2, opened 2018-11-12
Solution in Python

Ismael-Duran:


Python/chapter01/1.1 - Is Unique/Q1.1.py
```diff
@@ -0,0 +1,27 @@
...
+# implies a lowercase letter and the uppercase of
+# that letter is not considered duplicates
+# example: the string "Hather" has an uppercase "H"
+#          and lowercase "h" yet the string has no duplicate
+#          "h" or "H"
+
+
+def uniqueChar(passedStr):
+
+        # Will use the built-in function set() which
```
lhchavez: clarification, [`set`](https://docs.python.org/3/library/stdtypes.html#set) is a class/type, not a function. you're just invoking the class' constructor here.

Python/chapter01/1.1 - Is Unique/Q1.1.py
```diff
@@ -0,0 +1,27 @@
...
+        # eliminates dupplicates
+        deleteDuplicateIfAny = set(passedStr)
+
+        # If the string had unique characters the string
+        # length remains the same
+        # If the string has duplicates then size 
+        if len(deleteDuplicateIfAny) == len(passedStr):
```
lhchavez: it might be better to return a boolean instead of a string, since that is more testable:

```python3
return len(deleteDuplicatesIfAny) == len(passedStr)
```

Python/chapter01/1.1 - Is Unique/Q1.1.py
```diff
@@ -0,0 +1,27 @@
...
+                return "Unique"
+        else:
+                return "Not Unique"
+
+
+example = "Helloh"
+
+print(uniqueChar(example))
```
lhchavez: bonus points if you use [`unittest`](https://docs.python.org/3/library/unittest.html).

Python/chapter01/1.2 - Check Perm/Q1.2.py
```diff
@@ -0,0 +1,18 @@
+example = "hello"
+example2 = "ehloe"
+
+for letter in example:
+    if letter in example2:
+            index = example2.index(letter)
+            example2 = example2[:index] + example2[index+1:]
```
lhchavez: what is the algorithmic complexity of this loop? can you do better?

PR #3, opened 2018-11-12
python solution to 1.1 and 1.2

lomeli-noe:


Python/chapter01/1.1 - Is Unique/lomeli_1.1.py
```diff
@@ -0,0 +1,30 @@
...
+# What if you cannot use additional data structures.
+
+# For this problem I will assume we are working ASCII characters for which there
+# are 128 unique values.
+
+import array
+
+def isUnique(str):
+
+    # if the string is longer than 128 characters, it means there are duplicates
```
lhchavez: good use of the [Pigeonhole principle](https://en.wikipedia.org/wiki/Pigeonhole_principle), but this alienates the vast majority of the world (which doesn't use ASCII) :( do you think you can solve this by using a different data structure?

lomeli-noe: I think this would take care of the problem.

# Turn unicode string to plain ASCII
  str =  ''.join(c for c in unicodedata.normalize('NFD', str) if unicodedata.category(c)!= 'Mn')

lhchavez: that depends on the locale: in Spanish it might be okay, but this normalization violates some [language rules](http://userguide.icu-project.org/collation). it would also completely discard strings in non-latin-alphabet languages.

PR #11, opened 2018-12-18
Add chapter 1 solutions to 1.2 - 1.6

rroque98:


JavaScript/chapter01/1.3 - URLify/rroque98_sol.js
```diff
@@ -0,0 +1,8 @@
+// Write a method to replace all spaces within a string with '%20'
+
+const URLify = (string) => string.split(' ').join('%20');
```
lhchavez: you can also do

```javascript
const encodeSpaces = (string) => string.replace(/ /g, '%20');

File: JavaScript/chapter01/1.4 - PalinPerm/rroque98_sol.js [hunk 2]
```diff
@@ -0,0 +1,28 @@
...
+  const obj = {};
+  var oddCount = 0;
+  for (let char of strNoSpaces) {
+    if (obj[char] !== undefined) {
+      obj[char]++;
+    } else {
+      obj[char] = 1;
+    }
+  }
+  for (let key in obj) {
```
lhchavez: since you are not using the key for anything, it might be better to do

```javascript
const frequencies = new Map();
// ...
for (let frequency of frequencies.values()) {
  if (frequency % 2 == 0) {
    continue;
  }
  oddCount++;
  if (oddCount > 1) {
    return false;
  }
}
```

JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
```diff
@@ -0,0 +1,43 @@
...
+1 or 0 edits away*/
+
+const isOneAway = (str1, str2) => {
+  if (str1.length === str2.length) {
+    let errorCount = 0;
+    for (let i = 0; i < str1.length; i++) {
+      if (str1[i] !== str2[i]) {
+        errorCount++;
+      }
+      if (errorCount > 1) {
```
lhchavez: you don't need to check this unless you actually modify the error count. so you can use the [guard clause pattern](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) and reduce nesting while you're at it:

```javascript
for (let i = 0; i < str1.length; i++) {
  if (str1[i] === str2[i]) {
    continue;
  }
  errorCount++;
  if (errorCount > 1) {
    return false;
  }
}
```

JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
```diff
@@ -0,0 +1,43 @@
...
+        x++;
+      }
+      if (errorCount > 1) {
+        return false;
+      }
+      x++;
+    }
+    // ***** Helper functions ********
+    function findLongestStr(str1, str2) {
+      str1 > str2 ? str1 : str2;
```
lhchavez: what happens with `findLongestStr('ab', 'b')`? also, it might be better to be explicit and add the `return`.

JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
```diff
@@ -0,0 +1,43 @@
...
+      }
+      if (errorCount > 1) {
+        return false;
+      }
+    }
+  } else {
+    let errorCount = 0;
+    const longestStr = findLongestStr(str1, str2);
+    let x = 0;
+    for(let i = 0; i < longestStr; i++) {
```
lhchavez: this loop assumes that the lengths of the strings differe by at most 1, but it's not necessarily true. it might be better to actually check that explicitly.

JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
```diff
@@ -0,0 +1,43 @@
...
+      if (str1[i] !== str2[x]) {
```
lhchavez: what happens if the parameters are swapped?

JavaScript/chapter01/1.6 - String Compression/rroque98_sol.js
```diff
@@ -0,0 +1,33 @@
...
+    }
+  }
+  return compStr;
+}
+
+// TESTS
+console.log(stringCompression('aabcccccaaa') === 'a2b1c5a3');
+console.log(stringCompression('cccccccc') === 'c8');
+console.log(stringCompression('') === '');
+console.log(stringCompression('AabccCccaaa') === 'A1a1b1c2C1c2a3');
```
lhchavez:
```javascript
console.log(stringCompression('x') == 'x');
```

JavaScript/chapter01/1.2 - Check Perm/rroque98_sol.js
```diff
@@ -0,0 +1,37 @@
...
+  const obj2 = determineCharCount(str2);
+  for (let char of str1) {
+    if (obj1[char] !== obj2[char]) {
+      return false;
+    }
+  }
+  return true;
+  //******** helper function ******
+  function determineCharCount(string) {
+    const obj = {};
```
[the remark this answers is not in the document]
rroque98: Thanks for the feedback! Is this what you mean?

```suggestion
    function determineCharCount(string) {
        return string.split('').reduce((acc, char) => {
            if (acc[char] !== undefined) {
                acc[char]++;
            } else {
                acc[char] = 1;
            }
            return acc;
        }, {});
    }
```

JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
rroque98: Ok, I see what you mean. Thanks for the feedback

JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
rroque98: Oh yes, I meant to compare string lengths and have corrected this now. Thanks for catching this.

JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
rroque98: Here, I also meant to loop through longestStr.length. I must have been a bit tired when I worked on this problem. Thanks!

JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
rroque98: The logic won't work for all scenarios. I have changed my logic and will submit a pull request soon with the changes.

JavaScript/chapter01/1.6 - String Compression/rroque98_sol.js
rroque98: Good catch. I corrected an error with my logic and will submit in new PR

PR #13, opened 2019-01-10
Adding 2.1,2.2,2.3

camilosalazar98:


Python/chapter01/1.8 - Zero Matrix/camilo_solution_1.8.py
```diff
@@ -2,21 +2,67 @@
...
 
 
 def zeroMatrix(matrix):
-    m = len(matrix) - 1
-    n = len(matrix[0]) - 1
-    for i in range(m):
-        for j in range(n):
-                if matrix[i][j]  == 0:
-                    matrix[i+1][j] = 0
+    rowZero = False
```
lhchavez: how would you solve it if instead of a matrix, you had to do it with a single array? i'm betting something like:

```python
hasZero = False
for value in array:
  if value == 0:
    hasZero = True
    break
# sidenote, the above can be expressed as
#   hasZero = any(value == 0 for value in array)
if hasZero:
  for i in len(array):
    array[i] = 0
```

since matrices are just a bunch of arrays stacked together, you should be able to repeat the above snippet for every row and column (with two separate `for` loops). so two loops with a snippet of code _very_ similar to the one above should suffice to solve this problem.

Python/chapter01/1.9 - String Rotation/camilo_solution_1.9.py
```diff
@@ -0,0 +1,27 @@
+"""
+String Rotation:Assume you have a method isSubstring which checks if one word is a
+substring of another. Given two strings, sl and s2, write code to check if s2 is
+a rotation of sl using only one call to isSubstring (e.g.,"waterbottle" is a
+rotation of"erbottlewat").
+"""
+
+def isSubstring(s1,s2):
+    if s2 in s1: #check to see if the seqence of s2 is found in s1
```
lhchavez:
```python
return s2 in s1
```

suffices and you avoid the most common [antipattern](https://ucsb-cs16.github.io/topics/antipatterns/).

Python/chapter02/2.1 - Remove Dups/camilo_solution_2.1.py
```diff
@@ -0,0 +1,103 @@
...
+        self.head = None#clear the orignal link list
+        for i in range(len(d_)):
+            self.push(d_[i])#Use the element in the set a build a linklist with no dups
+
+
+
+
+
+
+#Testing to see if my class works
```
lhchavez: bonus points if you use the [`unittest`](https://docs.python.org/3/library/unittest.html) module.

camilosalazar98: Yeah, I really need to learn how to use unittest module.

lhchavez: it's pretty easy! just copy the boilerplate https://docs.python.org/3/library/unittest.html#basic-example and instead of printing, use `self.assertEqual` and you're done!

PR #15, opened 2019-01-16
Chapter3

nickolasteixeira:


Python/chapter03/3.6 - Animal Shelter/Nick.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def printQueue(self):
+        print('----- Beginning of Queue ------')
+        for x in self.queue[::-1]:
+            print(x)
+        print('------ End of Queue ------')
+
+    def dequeueAnimal(self, animal):
+        if len(self.queue) == 0:
+            return
```
lhchavez: this case is unnecessary, since it'll be handled naturally by the `for`.

Python/chapter03/3.6 - Animal Shelter/Nick.py
```diff
@@ -0,0 +1,88 @@
...
+        temp = False
+        for x in self.queue[::-1]:
```
lhchavez: here you can use the [guard clause pattern](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) instead of moving the condition to a different block:

```python3
for pet in self.queue[::-1]:
  if not pet.startswith(animal_type):
    continue
  self.queue.remove(pet)
  return pet
return None
```

Python/chapter03/3.6 - Animal Shelter/Nick.py
```diff
@@ -0,0 +1,88 @@
...
+Hints: #22, #56, #63'''
+import unittest
+
+
+class ShelterQueue:
+    def __init__(self):
+        self.queue = []
+
+    def enqueue(self, animal):
+        if len(self.queue) == 0:
```
lhchavez: this branch is unneded. the whole method can be just

```python3
self.queue.insert(0, animal)
```

Python/chapter03/3.6 - Animal Shelter/Nick.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def dequeueDog(self):
+        return self.dequeueAnimal('D')
+
+    def dequeueCat(self):
+        return self.dequeueAnimal('C')
+
+    def dequeueAny(self):
+        if len(self.queue) == 0:
+            return
```
lhchavez: [explicit is better than implicit](https://www.python.org/dev/peps/pep-0020/#id3)

```python3
return None
```

Python/chapter03/3.4 - Queue via Stacks/Nick.py
```diff
@@ -0,0 +1,123 @@
...
+
+        if item == self.min:
+            if self.length() > 0:
+                self.findMin()
+            else:
+                self.min = None
+        return item
+
+    def findMin(self):
+        self.min = self.stack[0]
```
lhchavez: you can just do

```python3
self.min = min(self.stack)
```

Python/chapter03/3.3 - Stack of Plates/Nick.py
```diff
@@ -0,0 +1,151 @@
...
+        self.min = self.stack[0]
+        for x in self.stack:
+            if x < self.min:
+                self.min = x
+
+    def length(self):
+        return len(self.stack)
+
+    def push(self, value):
+        if not self.min:
```
lhchavez: are you guaranteed that there will be no zeroes in the input data? otherwise this might misfire with zeroes. a more robust check is

```python3
if self.min is None or self.min > value:
  self.min = value
```

PR #16, opened 2019-01-18
logic change for permutations - chapter 1

nickolasteixeira:


Python/chapter01/1.4 - PalinPerm/Nick4.py
```diff
@@ -15,15 +15,15 @@ def palperm(string):
         string (str): a string to check for all palindrome permutations
     '''
     table = {}
-    count = 0
     for letter in string.lower().replace(" ", ""):
         table[letter] = table.get(letter, 0) + 1
 
+    count = 0
```
lhchavez: might be better to use [`dict.values()`](https://docs.python.org/3/library/stdtypes.html#dict.values):

```suggestion
    for value in table.values();
      if value % 2 == 1:
          return False
    return True
```

that can also be expressed in a one-liner with [`all()`](https://docs.python.org/3/library/functions.html#all):

```suggestion
    return all(value % 2 != 1 for value in table.values())
```

Python/chapter01/1.4 - PalinPerm/Nick4.py
```diff
@@ -15,15 +15,15 @@ def palperm(string):
         string (str): a string to check for all palindrome permutations
     '''
     table = {}
```
lhchavez: this can be expressed more concisely with [`collections.defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict), and give it a more self-explanatory name while we're at it:

```suggestion
     letter_frequencies = collections.defaultdict(int)
     for letter in string.lower().replace(' ', ''):
       letter_frequencies[letter] += 1
```

darrylbalderas:
https://github.com/techqueria/data-and-algorithms/blob/329a2f2db89ece312e6462b73259d711e173d4d1/Python/chapter01/1.4%20-%20PalinPerm/Nick4.py#L17-L19

Since you are using the collection module, you could use Counter to the letters 

```
table  = collections.Counter(string.lower().replace(" ",""))
```

Python/chapter01/1.4 - PalinPerm/Nick4.py
darrylbalderas: @lhchavez Does also following follow the guard clause pattern ?

```suggestion
    odd_frequency_count = 0
    for frequency in letter_frequencies.values:
        if frequency % 2 == 1:
            odd_frequency_count += 1
            if odd_frequency_count > 1:
             return False
    return True
```

lhchavez: you'd need to flip the condition:

```python
for frequency in letter_frequencies.values:
  if frequency % 2 == 0:
    continue
  odd_frequency_count += 1
  if odd_frequency_count > 1:
    return False
return True
```

an easy to remember heuristic is to try to minimize the indentation level. the real guidance is that the `if ` blocks (a.k.a. the guard clauses) should cover the non-main case and keep that logic out from the most important path. that way it's easy to recognize the different parts of the algorithm just based on visual inspection.

fvntr:
ready to merge?

nickolasteixeira:
Yes, I included a new function per @lhchavez's feedback.

PR #17, opened 2019-01-21
Miguel solution to 1.1 [Python]

miguelHx:
I would appreciate any feedback, especially critical feedback :)


Python/chapter01/1.1 - Is Unique/miguel_1.1_sol.py
```diff
@@ -0,0 +1,60 @@
...
+california	false
+orbit		true
+e		    true
+"""
+
+
+def is_unique(input_str):
+    freqs = {}
+    for c in input_str:
+        if c not in freqs:
```
lhchavez: you are treating `freqs` as a set and not a map, so how about you make it a legit [`set`](https://docs.python.org/3/library/stdtypes.html#set) object?

Python/chapter01/1.1 - Is Unique/miguel_1.1_sol.py
```diff
@@ -0,0 +1,60 @@
...
+            freqs[c] = 1
+        else:
```
lhchavez: you can use the [guard clause pattern](http://wiki.c2.com/?GuardClause) here to avoid over-indenting:

```python3
if c in freqs:
  return False
freqs.add(c)
```

Python/chapter01/1.1 - Is Unique/miguel_1.1_sol.py
```diff
@@ -0,0 +1,60 @@
...
+            return False
+    return True
+
+
+def is_unique_no_additional_data_structures(input_str):
+    for i in range(0, len(input_str)):
```
lhchavez: since you want both the index and the element, you can use [`enumerate()`](https://docs.python.org/3/library/functions.html#enumerate):

```suggestion
    for i, c in enumerate(input_str):
```

Python/chapter01/1.1 - Is Unique/miguel_1.1_sol.py
```diff
@@ -0,0 +1,60 @@
...
+        for j in range(i+1, len(input_str)):
```
lhchavez: here you just want to check if the character is contained within a substring:

```suggestion
        if c in input_str[i+1:]:
            return False
```

Python/chapter01/1.1 - Is Unique/miguel_1.1_sol.py
```diff
@@ -0,0 +1,60 @@
+"""
+Python version 3.7.0
+1.1
+Implement an algorithm to determine if a string has all unique characters.
+What if you cannot use additional data structures?
```
lhchavez: one thing you should ask (and be able to argue an answer): what about the empty string?

miguelHx: Going to push my latest code addressing your review, thank you for the feedback! 👍

Python/chapter01/1.1 - Is Unique/miguel_1.1_sol.py
```diff
@@ -0,0 +1,101 @@
...
+    if input_str == "":
+        raise ValueError("empty input string")
+
+
+def is_unique(input_str):
+    """
+    Determines if input_str has all unique characters.
+    input string MUST have at least 1 character. n>=1, where n is the number of characters in a string.
+    Hence, the domain is now defined. Any input not within the domain will not be considered. (n < 1)
+    An exception will be raised to handle faulty input.
```
lhchavez: it's a good practice to define what "faulty input" means in the docstring for your public interface. This is formally known as [preconditions](https://en.wikipedia.org/wiki/Precondition).

Python/chapter01/1.1 - Is Unique/miguel_1.1_sol.py
```diff
@@ -0,0 +1,101 @@
...
+    for c in input_str:
+        if c in chars_seen:
+            return False
+        chars_seen.add(c)
+    return True
+
+
+def is_unique_no_additional_data_structures(input_str):
+    """
+    Variant of is_unique.  Uses no additional data structures (besides the variables from the iterator)
```
lhchavez: If someone was just reading this docstring, they would probably think this method is the best to use ("hey! no additional memory!", they might think), so if there are any tradeoffs, it might be good to explicitly call them out.

Python/chapter01/1.1 - Is Unique/miguel_1.1_sol.py
miguelHx: Looks like the way `all([])` was defined like a universal quantifier from mathematical logic.  It is promised that for all X in S, X is True.  According to https://en.wikipedia.org/wiki/Universal_quantification#The_empty_set
the universal quantifier of an empty set is defined to be true by convention.  Hence why all([]) returns true.  
In mathematical logic, it makes sense that all([]) returns true because we are following the mathematical rules.  
However, in the case of is_unique, I guess I could have the function return true for an empty string, because the problem can be expressed as a universal quantifier (for all c in input_str, there are no two c's that are equal).  But for practical purposes, I think that it's best to let the user know that an empty string was passed in, and they can handle however they wish with a try-catch block. 
What are your thoughts on that?
Source:  https://stackoverflow.com/questions/11979683/why-python-built-in-all-function-returns-true-for-empty-iterables

lhchavez: if you are pushing design decisions on the user (e.g. forcing them to catch an exception), the effort required to check for the empty string before they pass it in as an argument is roughly equivalent.

I *personally* prefer to be consistent with the universal qualifier interpretation since it also allows for slightly simpler implementation since you can just ignore the "special case" of the empty string and the code will naturally reach the `return True` (consistency ftw!), but in reality both solutions are correct, as long as you clearly document the precondition.

Python/chapter01/1.1 - Is Unique/miguel_1.1_sol.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    _validate_input_str(input_str)
+    for i, c in enumerate(input_str):
+        if c in input_str[i+1:]:
+            return False
+    return True
+
+
+class TestIsUniqueFunction(unittest.TestCase):
+    def _run_tests(self, f: callable([[str], None])) -> None:
```
lhchavez: this needs to be `Callable` and you also need to [`from typing import Callable`](https://docs.python.org/3/library/typing.html#typing.Callable).

Python/chapter01/1.1 - Is Unique/miguel_1.1_sol.py
miguelHx: After the user runs into the exception, they can just check for an empty string next time rather than using a try-catch block, so it is forced only if they desire to use the empty string.  So I think it will serve more as a warning, rather than a try catch override.  But then again, it probably depends on the client programmer.

To what point should the input validation be left to the user of the library? (beyond checking for the input type).  Perhaps it should stay flexible.  The user will be guided by the documented precondition, as you have stated.

After digging a little further, I've found that the statement that asserts that all members of an empty set have a certain property is known as the Vacuous Truth.
According to https://en.wikipedia.org/wiki/Vacuous_truth
From the wikipedia link:
> "Outside of mathematics, statements which can be characterized informally as vacuously true can be misleading.  Such statements make reasonable assertions about qualified objects which do not actually exist. For example, a child might tell their parent 'I ate every vegetable on my plate', when there were no vegetables on the child's plate to begin with."

The quote references to statements that are "informally" characterized.  That is not the case here, since we are formally defining the problem.  However, not every programmer (I assume) has a background in formal mathematical logic, so I can see from their perspective why empty-set results as we have been discussing can be misleading.

I understand, though,  that it comes down to personal preference (as you have hinted), and that both approaches are correct.

Python/chapter01/1.1 - Is Unique/miguel_1.1_sol.py
miguelHx: added a note for runtime drawback

Python/chapter01/1.1 - Is Unique/miguel_1.1_sol.py
miguelHx: Oh ok, will update that

Python/chapter01/1.1 - Is Unique/miguel_1.1_sol.py
miguelHx: I'm gonna go with your suggestion tho.  Pushing latest code in a few.

lhchavez: note that "after the user runs into the exception" should be something you should actively try to design against. this is called the [Principle of Least Astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

being consistent with the rest of the ecosystem, following the surrounding codebase's style, and relying on analogues like the universal qualifier argument are all tools to help this.

> To what point should the input validation be left to the user of the library?

**tl;dr** it depends.

normally the libraries perform validation needed to assert that the preconditions are followed and to give better error diagnostics. for instance, null-checks are rarely performed by libraries since a) you can add type annotations in Java and Python that let the user / compiler know that they shouldn't be passing around null values, and b) the exception thrown by failure to comply with the precondition will be very obvious to the caller. on the other hand, if you have a more complex precondition (say, you are doing an invert matrix routine and the user passes in a [singular matrix](https://en.wikipedia.org/wiki/Invertible_matrix)), unless you do an active validation, the user will be left with a very confusing stack trace when an exception happens in the middle of the function with no obvious reason. if your language and compiler supports helping you with validation in _any_ way (such as type annotations), by all means take advantage of it, since having a compile-time error is much better than having runtime errors.

in the end, everything is a balancing act. you and your team should assign some vague value to the robustness, performance, maintainability, and ergonomics of the codebase. you _can_ write code that meets 100% in all categories (most of the time), but that would consume time dedicated to writing actual business logic. then again, if you _only_ focus on features business logic and neglect everything else, you most likely will end up with an unmaintainable spaghetti code mess that's awful to use and unexplainably slow.

miguelHx: Noted. 👍 
Thank you, once again, for your feedback.  This was all incredibly helpful.

PR #18, opened 2019-01-23
Miguel solution to 1.2 - Check Permutation [Python]

miguelHx:


Python/chapter01/1.2 - Check Perm/miguel_1.2_sol.py
```diff
@@ -0,0 +1,83 @@
...
+    "", ""          True
+    "test2", ""         False
+    :param s1: string of size n
+    :param s2: string of size m
+    :return: True if s1 is a permutation of s2, False otherwise
+    """
+    # permutations must be of equal length, one-to-one
+    if len(s1) != len(s2):
+        return False
+    freqs_s1 = {}
```
lhchavez: you can use a [`collections.Counter`](https://docs.python.org/3/library/collections.html#collections.Counter) for this.

Python/chapter01/1.2 - Check Perm/miguel_1.2_sol.py
```diff
@@ -0,0 +1,83 @@
...
+    # build histogram of seen characters in s1
+    # using histogram because string could have repeated characters
+    for c in s1:
+        if c not in freqs_s1:
```
lhchavez: this check-if-exists-otherwise-mutate pattern can be avoided with [`collections.defaultdict(int)`](https://docs.python.org/3/library/collections.html#collections.defaultdict)

Python/chapter01/1.2 - Check Perm/miguel_1.2_sol.py
```diff
@@ -0,0 +1,83 @@
...
+            freqs_s1[c] = 1
+            continue
+        freqs_s1[c] += 1
+
+    freqs_s2 = {}
+    for c in s2:
```
lhchavez: if you find yourself repeating almost the same code, consider moving the common code to a function and do an extra pass to execute the non-common code. it's more explicit that way and you might avoid common errors like the [last line effect](https://www.viva64.com/en/b/0260/)

Python/chapter01/1.2 - Check Perm/miguel_1.2_sol.py
miguelHx: I tried out `collections.defaultdict(int)` and I see the advantage of it, but ended up sticking with using solely `collections.Counter()` objects, as you recommended.

Python/chapter01/1.2 - Check Perm/miguel_1.2_sol.py
lhchavez: your assumption is correct!

there's always a trade-off between doing things eagerly vs. lazily, and there is high value for programmers to be lazy. You can read arguments in favor of delaying implementation in Martin Fowler's [YAGNI](https://martinfowler.com/bliki/Yagni.html) and against eager refactoring in Sandi Metz' [The Wrong Abstraction](https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction).

Now, you might notice that the last article sort of contradicts the initial suggestion. You need to strike the right balance, and you'll eventually get that with enough practice because both extremes of aggressively refactoring might end up creating tons of wrong abstractions, and duplicating _all the things_ risks introducing off-by-one bugs.

having said _all_ that, **if** you check that the lengths are equal, you can lose this block altogether! try to come up with a rough proof of why this works.

miguelHx: **PR updated** 
It seems that the following code block (call it code block B): 
```python
    # code block B
    for key, val in freqs_s1.items():
        if val != freqs_s2[key]:
            # character counts between s1 and s2 don't match
            return False
    return True
```
automatically takes care of a main part of that `are_characters_in_histogram` was checking for.  Namely, that for each c in s2, c occurs at least once in s1, and code block B will detect that as well.  So `are_chars_in_hist` is redundant, for the most part.

**However**, the ONE part that code block B does NOT account for is the case where s1 is an _empty_ string and s2 contains at least 1 character.  `are_characters_in_histogram` DOES account for that, because we are looping through characters in s2 to see if they exist in s1, which will correctly return False.

So code block B cannot _fully_ replace `are_characters_in_histogram` functionality, because code block B (without `are_chars_in_hist` and without the length check) will incorrectly return True in the case where s1 is empty and s2 contains at least one character.

If I take out the block `are_chars_in_histogram` AND the if statement checking the lengths, the program will work for the most part except in the case where we have s1 as an empty string and s2 with at least one char.

Checking that the lengths are equal will allow us to lose the block `are_chars_in_hist` because the length check will account for detecting the empty s1 string (s2 has at least 1 char) scenario.

Python/chapter01/1.2 - Check Perm/miguel_1.2_sol.py
miguelHx: Code now uses collections.Counter 👍

Python/chapter01/1.2 - Check Perm/miguel_1.2_sol.py
lhchavez: you can generalize your argument by saying that the block won't detect cases where s1 is a permuted substring of s2, and every extra character was not found in s1 (e.g. "hola" and "holaxxxyyyzzz"). checking the length beforehand will prevent this from happening because:

* assume the second loop never caught the mistake and accidentally returned True.
* that would mean that every single character in s1 was contained in s2, but then the lengths are equal, so s2 could not have any extra characters.
* this is a contradiction, so the second block + length check was correct.

miguelHx: Ah that is clever and succinct.  Thanks!

Python/chapter01/1.3 - URLify/miguel_1.3_sol.py
```diff
@@ -0,0 +1,52 @@
...
+    :param s: the original string to 'urlify'
+    :param true_length: since s may have additional characters, we focus on true_length instead of actual s length
+    :return: string with each space from s replaced with '%20'
+    """
+    output = ""
+    for i, c in enumerate(s):
+        if i == true_length:
+            break
+        if c == ' ':
+            output += "%20"
```
lhchavez: what is the runtime complexity of this algorithm?

Python/chapter01/1.3 - URLify/miguel_1.3_sol.py
```diff
@@ -0,0 +1,52 @@
...
+            continue
+        output += c
+    return output
+
+
+class TestUrlifyFunction(unittest.TestCase):
+    def _run_tests(self, f: Callable[[str, int], str]) -> None:
+        cases = [
+            ("Mr John Smith     ", 13, "Mr%20John%20Smith"),
```
lhchavez: here you can do something like

```python3
cases = [
  (("Mr John Smith     ", 13), "Mr%20John%20Smith"),
  ...
]
for args, expected in cases:
  self.assertEqual(urlify(*args), expected, msg=...)
```

no need to pass in `urlify` since that was just for the other test where you had multiple implementations invoking the same test cases.

lhchavez: approved but don't have merge powers here. we'll need to wait for @fvntr

fvntr:
good to merge?

miguelHx:
@fvntr yes 👍

PR #19, opened 2019-01-26
Miguel 1.3 - URLify [Python]

miguelHx:


Python/chapter01/1.3 - URLify/miguel_1.3_sol.py
```diff
@@ -0,0 +1,52 @@
...
+    :param s: the original string to 'urlify'
+    :param true_length: since s may have additional characters, we focus on true_length instead of actual s length
+    :return: string with each space from s replaced with '%20'
+    """
+    output = ""
+    for i, c in enumerate(s):
+        if i == true_length:
+            break
+        if c == ' ':
+            output += "%20"
```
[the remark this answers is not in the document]
miguelHx: O(true_length), since the number of steps this code takes depends on true_length.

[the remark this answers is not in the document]
lhchavez: rephrasing the question: what's the runtime complexity of one string concatenation in Python?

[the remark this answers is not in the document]
miguelHx: O(n^2) since the string will be rebuilt at each step.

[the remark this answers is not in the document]
miguelHx: Looking online, looks like .join() might be better.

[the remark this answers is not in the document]
miguelHx: Oh whoops, shouldn't the runtime complexity of one string concatenation be O(n), where n is the length of the string?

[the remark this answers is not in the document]
lhchavez: correct: O(n) for each string concatenation, and you'll be doing O(n) string concatenations, so total will be O(n²).

Python/chapter01/1.3 - URLify/miguel_1.3_sol.py
```diff
@@ -0,0 +1,54 @@
...
+    - length of s <= true_length
+    :param s: the original string to 'urlify'
+    :param true_length: since s may have additional characters, we focus on true_length instead of actual s length
+    :return: string with each space from s replaced with '%20'
+    """
+    # Below link goes over string concat efficiency in python (I will use method 4
+    # https://waymoot.org/home/python_string/
+    output = []
+    for i, c in enumerate(s):
+        if i == true_length:
```
lhchavez: note that this will almost never be true! if `true_length` is equal to `len(s)`, `i` is at most `len(s) - 1`.

the reason why `true_length` is provided is that this assumes that you are coding in C/C++ and you will be modifying the string in-place, but this is not true in Java or Python (hence the weird comment about the character array thing).

if you still want to honor it (e.g. for the case where you have an "ignore this" suffix), do you think you can do it without introducing an additional branch?

miguelHx: Oh ok, I see.  I'll add another function without true_length.  Then I will explain in the docstring why true_length is omitted.

lhchavez: sure, and if you do end up keeping true_length, try to think of a way to _not_ have an explicit `if i == true_length` check. it's possible and Pythonic.

miguelHx: Oh just saw this on refresh, will attempt pythonic way for that now

miguelHx:
```python
for c in s[0:true_length]
```

Python/chapter01/1.3 - URLify/miguel_1.3_sol.py
```diff
@@ -0,0 +1,87 @@
...
+            ("Mr John Smith", "Mr%20John%20Smith"),
+            ("Miguel Hernandez", "Miguel%20Hernandez"),
+            (" Techqueria ", "%20Techqueria%20"),
+            ("a b c d e f g h", "a%20b%20c%20d%20e%20f%20g%20h"),
+            ("a b c d e f g h ignore this", "a%20b%20c%20d%20e%20f%20g%20h%20ignore%20this"),
+            ("ihavenospaces", "ihavenospaces"),
+            ("nospacesIgnoreme", "nospacesIgnoreme")
+        ]
+        for args, expected in cases:
+            self.assertEqual(f1(*args), expected, msg=args)
```
lhchavez: this utility function did make sense for the previous problem since you were calling one set of tests for two different functions, but now it doesn't since you're calling two sets of tests for two different functions.

so how about you remove this function altogether, call `urlify` and `urlify_no_true_length` directly (to prevent having the reader to to the callsite to figure out what `f1` and `f2` actually mean), and finally splitting the test method in two: one for `urlify` and another for `urlify_no_true_length`?

Python/chapter01/1.3 - URLify/miguel_1.3_sol.py
```diff
@@ -0,0 +1,87 @@
...
+    Precondition(s):
+    - length of s <= true_length
+    :param s: the original string to 'urlify'
+    :param true_length: since s may have additional characters, we focus on true_length instead of actual s length
+    :return: string with each space from s replaced with '%20'
+    """
+    # Below link goes over string concat efficiency in python (I will use method 4
+    # https://waymoot.org/home/python_string/
+    output = []
+    for c in s[0:true_length]:
```
lhchavez: optional: you already have a solution in the Pythonic way (urlify_no_true_length), but if you want to do the exercise as the author of the book intended, try solving it in under the following constraints:

```python3
result = ['\x00'] * true_length
for i, c in s:
  result[i] = c
# your code. you are only allowed to allocate O(1) extra memory.
return ''.join(result).rstrip('\x00')
```

miguelHx: Gave it an attempt.  My attempt was not as clean as the pythonic way, but it was fun to try it out.  It somehow felt odd to code in python as if it was java or c/c++ lol.  But I also feel that it was good practice.

I changed the result length from true_length to len(s) * 2, since for each character up to true_length in s, the size of result will increase by 2 characters.

Python/chapter01/1.3 - URLify/miguel_1.3_sol.py
```diff
@@ -0,0 +1,96 @@
...
+    Precondition(s):
+    - length of s <= true_length
+    :param s: the original string to 'urlify'
+    :param true_length: since s may have additional characters, we focus on true_length instead of actual s length
+    :return: string with each space from s replaced with '%20'
+    """
+    # Below link goes over string concat efficiency in python (I will use method 4
+    # https://waymoot.org/home/python_string/
+    # each space will make us replace it (with '%', then add 2 characters ('2', '0')
+    result = ['\x00'] * len(s)*2  # extra buffer space
```
lhchavez: for this challenge to be successful, you cannot use extra buffer space:

```suggestion
    result = ['\x00'] * true_length
```

the contract in C/C++ is that you **must** tell library functions the truth: if the function says you need to specify the length of the buffer in a way that the final result fits, you should trust that. any crashes that happen because you were lied to are the caller's fault, not yours.

in any case, if you pass a string with a single space, your version would crash since it needs 3, not 2 slots.

Python/chapter01/1.3 - URLify/miguel_1.3_sol.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    Given a string and it's "true" length, this function will return a new string
+    that replaces all of the spaces of the input string with '%20'.
+    Precondition(s):
+    - length of s <= true_length
+    :param s: the original string to 'urlify'
+    :param true_length: since s may have additional characters, we focus on true_length instead of actual s length
+    :return: string with each space from s replaced with '%20'
+    """
+    # Below link goes over string concat efficiency in python (I will use method 4
```
lhchavez: this comment is now outdated.

Python/chapter01/1.3 - URLify/miguel_1.3_sol.py
```diff
@@ -0,0 +1,96 @@
...
+    # https://waymoot.org/home/python_string/
+    # each space will make us replace it (with '%', then add 2 characters ('2', '0')
+    result = ['\x00'] * len(s)*2  # extra buffer space
+    i = 0
+    space_count = 0
+    added_chars_per_space = 2
+    for c in s:
```
lhchavez: i forgot to add a restriction to the challenge: you cannot reference `s`: in C/C++, you wouldn't have that luxury.

Python/chapter01/1.3 - URLify/miguel_1.3_sol.py
```diff
@@ -0,0 +1,96 @@
...
+        if i == true_length + added_chars_per_space * space_count:
```
lhchavez: try to write the loop in a way that you don't need this index-check.

Python/chapter01/1.3 - URLify/miguel_1.3_sol.py
miguelHx: For the final result to fit, shouldn't the size be true_length + 2*space_count?

Going to add an extra test case for single space.

miguelHx: In the example input/output:
input:  "Mr John Smith    ", 13
output:  "Mr%20John%20Smith"

let's assume that true_length tells us what the output length should be.

Based on the example, that would mean that the output string must be of size 13.
The length of the output is 17, we have a contradiction.

So true_length is not the length of what the output should be (according to the author, at least)

Or maybe it could be that the author made a typo.

Python/chapter01/1.3 - URLify/miguel_1.3_sol.py
```diff
@@ -24,22 +24,21 @@ def urlify(s: str, true_length: int) -> str:
...
-            break
-        if c == ' ':
-            result[i] = '%'
-            result[i+1] = '2'
-            result[i+2] = '0'
-            i += 3
+    chars_per_space = 2
+    idx = 0
+    for x in range(true_length):
+        if s[idx - space_count * chars_per_space] == ' ':
```
lhchavez: remember that `s` is inaccessible in the challenge mode.

miguelHx: I'm not sure I understand.  How else would I get the character values of the argument s?

lhchavez: actually let me formalize the challenge to avoid confusion:

```python3
from typing import List

def urlify_challenge(s: str) -> str:
  """In-place version of urlify."""
  buf = ['\x00'] * (len(s) * 3)
  for i, c in enumerate(s):
    buf[i] = c
  def _challenge(buf: List[str], original_length: int) -> None:
    # Your code here. Challenge rules:
    # * You can only reference |buf| and |original_length|, not |s|.
    # * You can only allocate O(1) additional memory
    # * You need to modify |buf| in-place so that the right answer is returned.
  _challenge(buf, len(s))
  return ''.join(buf).rstrip('\x00')
```

you can get the character values through the buffer.

miguelHx: Oh ok, I see what you mean now.  Will give that a go:)

miguelHx: Lol I mistakenly still used s in my initial _challenge attempt, idk how that missed that.  My negligence.

Has been updated to modify buf in place and NOT reference s.

PR #21, opened 2019-01-28
Add JavaScript solution to 1.7, 1.8, 2.1, 2.3 and refactor 1.2 - 1.6

rroque98:


JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
```diff
@@ -18,19 +19,32 @@ const isOneAway = (str1, str2) => {
     let errorCount = 0;
     const longestStr = findLongestStr(str1, str2);
     let x = 0;
-    for(let i = 0; i < longestStr; i++) {
-      if (str1[i] !== str2[x]) {
-        errorCount++;
+    for(let i = 0; i < longestStr.length; i++) {
+      let currentChar1 = str1[i];
+      let currentChar2 = str2[x]
+      let nextChar1 = str1[i + 1];
```
lhchavez: it's not completely obvious to a random person reading your code why `i+1` or `x+1` are safe (actually i'm not even sure myself).

A few alternatives:

a) modify the loop condition so that it takes this into account:
```javascript
for (let i = 0, x = 0; i + 1 < str1.length && x + 1 < str2.length; i++) {
```
b) Add a comment with a rough "proof" of why this will never ever crash.

JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
```diff
@@ -18,19 +19,32 @@ const isOneAway = (str1, str2) => {
...
+      if (str1[i] === str2[x]) {
         x++;
+        continue;
       }
+      errorCount++;
       if (errorCount > 1) {
         return false;
       }
-      x++;
+      if (currentChar1 === nextChar2) {
```
lhchavez: having the code handle both cases at the same time is hard to follow. do you think there could be a way in which you only need to deal with the case where both strings are the same length (block L7-L18) and `str1.length < str2.length`?

JavaScript/chapter01/1.6 - String Compression/rroque98_sol.js
```diff
@@ -8,9 +10,9 @@ const stringCompression = (str) => {
     return '';
   }
   var compStr = '';
-  var count = 1;
+  var count = 0;
```
lhchavez: all of these can be `let` instead of `var`.

JavaScript/chapter01/1.6 - String Compression/rroque98_sol.js
```diff
@@ -8,9 +10,9 @@ const stringCompression = (str) => {
...
   var currentChar = str[0];
-  for (let i = 1; i < str.length; i++) {
+  for (let i = 0; i < str.length; i++) {
     let char = str[i];
     if (char === currentChar) {
       count++;
```
lhchavez: I cannot add a comment on L19, but this is the well-known [for-if antipattern](https://blogs.msdn.microsoft.com/oldnewthing/20111227-00/?p=8793).

Do you think you can restructure the code in a way in which you avoid checking `if (i === str.length - 1)`?

JavaScript/chapter01/1.7 - Rotate Matrix/rroque98_sol.js
```diff
@@ -0,0 +1,25 @@
+/*Given an image represented by an N*N matrix where
```
lhchavez: good PR ettiquette dictates that you should try to avoid fixing stuff _and_ adding new things into the same PR:

* https://medium.com/@hugooodias/the-anatomy-of-a-perfect-pull-request-567382bb6067
* https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#separate-your-changes
* https://medium.com/@fagnerbrack/one-pull-request-one-concern-e84a27dfe9f1

do you mind moving this to a separate PR? that way if one PR gets a lot of comments and the other one does not, you still do net progress since you'll be able to land one faster.

JavaScript/chapter01/1.8 - Zero Matrix/rroque98_sol.js
```diff
@@ -0,0 +1,84 @@
...
+  [2, 0, 2],
+]
+
+var resultArr3 = [
+  [0, 0, 0],
+  [0, 0, 0],
+  [0, 0, 0],
+  [0, 0, 0],
+]
+console.log(JSON.stringify(zeroMatrix(testArr1)) === JSON.stringify(resultArr1));
```
lhchavez: it's significantly easier to understand if you bring the matrices [closer to the places where you are using them](http://wiki.c2.com/?DeclareVariablesAtFirstUse):

```suggestion
const testArr1 = [  // `var` is old hat, `const` is the new hotness!
  [3, 5, 6],
  [1, 0, 2],
  [4, 4, 5],
  [2, 2, 2],
]; // note the semicolons
const resultArr1 = [
  [3, 0, 6],
  [0, 0, 0],
  [4, 0, 5],
  [2, 0, 2],
];
console.log(JSON.stringify(zeroMatrix(testArr1)) === JSON.stringify(resultArr1));
```

JavaScript/chapter01/1.8 - Zero Matrix/rroque98_sol.js
```diff
@@ -0,0 +1,84 @@
+/*Write an algorithm such that if an element
+ in an M*N matrix is 0,it's entire row and column
+  are set to 0*/
+
+const zeroMatrix = (nestedArr) => {
+  var rowsAndColsWithZeros = checkForZeroIndex(nestedArr);
+  let rowLength = nestedArr[0].length;
+  let columnLength = nestedArr.length;
+  for (let i = 0; i < rowsAndColsWithZeros['rows'].length; i++) {
```
lhchavez: what's the runtime complexity of an NxN matrix of zeroes?

JavaScript/chapter01/1.8 - Zero Matrix/rroque98_sol.js
```diff
@@ -0,0 +1,84 @@
...
+    let rowIndex = rowsAndColsWithZeros['rows'][i];
+    for (let x = 0; x < rowLength; x++) {
+      nestedArr[rowIndex][x] = 0;
+    }
+  }
+  for (let i = 0; i < rowsAndColsWithZeros['columns'].length; i++) {
```
lhchavez: you don't actually use `i` except for indexing into the array, so you can use the more idiomatic

```javascript
for (const colIndex of rowsAndColsWithZeroes['columns']) {
```

Same above.

rroque98:
Apologies for the delayed response. I will break this pull request into smaller pull requests with the suggested feedback and will resubmit them soon. Thank you very much for the feedback.

rroque98:
I am going to close this pull request since I broke this PR into smaller pull requests already. Can someone with write access merge my other pull requests? They have already been approved. TY

PR #22, opened 2019-01-31
Miguel 1.5 - One Away [Python]

miguelHx:


Python/chapter01/1.5 - OneAway/miguel_1.5_sol.py
```diff
@@ -0,0 +1,67 @@
...
+            ("pale", "ple", True),
+            ("pales", "pale", True),
+            ("pale", "bale", True),
+            ("pale", "bake", False),
+            ("", "swag", False),
+            ("paale", "pale", True),
+            ("a", "", True),
+            ("", "b", True),
+            ("a", "b", True),
+            ("", "", True)
```
lhchavez: what about this case?

```suggestion
            ("", "", True),
            ("pale", "elap", False),
```

(note the trailing comma, that helps you keep adding to lists without having to touch the last entry).

Python/chapter01/1.5 - OneAway/miguel_1.5_sol.py
```diff
@@ -0,0 +1,67 @@
...
+        return False
+
+    char_freqs_s1 = collections.Counter(s1)
+    char_freqs_s2 = collections.Counter(s2)
+
+    anomalies = 0
+    for key, value in char_freqs_s1.items():
+        if key not in char_freqs_s2 or value != char_freqs_s2[key]:
+            anomalies += 1
+        if anomalies > 1:
```
lhchavez: nit: only check `anomalies` if you are modifying it (i.e. indent this block so that it only executes if L39 executes). there's no point in checking it over and over again!

Python/chapter01/1.5 - OneAway/miguel_1.5_sol.py
```diff
@@ -18,27 +17,43 @@ def one_away(s1: str, s2: str) -> bool:
...
-    for key, value in char_freqs_s1.items():
-        if key not in char_freqs_s2 or value != char_freqs_s2[key]:
-            anomalies += 1
-        if anomalies > 1:
-            return False
+    addend = 0
+    for i, c in enumerate(s_short):
+        if c == s_long[i + addend]:
+            continue
+        if i == i + addend:
```
lhchavez: this could be simplified to `if not added:`.

but then again, it might be even better to keep consistency with L36-38:

```python
added += 1
if added > 1:
  return False
```

Python/chapter01/1.5 - OneAway/miguel_1.5_sol.py
```diff
@@ -0,0 +1,85 @@
...
+            if edits > 1:
+                return False
+        return True
+    # otherwise, we have an insertion/deletion
+    # compare both strings for insertion/deletion
+    # want to loop through the shorter string to avoid going out of bounds
+    s_short = s1 if len(s1) < len(s2) else s2
+    s_long = s1 if len(s1) > len(s2) else s2
+
+    addend = 0
```
lhchavez: nit: maybe `added` (as in, "number of added characters"). while addend is technically correct (since you are adding it to `i`), it's too abstract. kinda like naming an integer variable `number`.

PR #25, opened 2019-02-01
Implement shell script for creating folder structure for new programming languages

darrylbalderas:
This PR is to address #24


createFolders.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+PROJECT_NAME=data-and-algorithms
+DIRECTORY=${PWD}
```
lhchavez: you can avoid most of the boilerplate and validation by using

```shell
DIRECTORY="$(readlink -f -- "$(dirname -- "$0")")"
```

yes, all those quotes are needed. otherwise you'll break people who clone this repository in a path that has spaces!

createFolders.sh
```diff
@@ -0,0 +1,18 @@
...
+   if [[("$BASE_DIRECTORY" != "$PROJECT_NAME")]]; then
+      echo "Make sure you are in $PROJECT_NAME project"
+      exit
+   fi
+   rootfileName=$1
+   mkdir ./$rootfileName
+   cp -a ./Python/ ./$rootfileName/
+   rm ./$rootfileName/*/*/*.py
+else
+   echo "Missing folder name: sh createFolders.sh <folder_name>"
```
lhchavez: you can use the [guard clause pattern](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html)

```
if [[ $# -lt 1 ]]; then
  echo "Usage: bash createFolders.sh <programming language>" > &2
  exit 1
fi
# The rest. You can now use $1 confidently.
```

the `>&2` prints to stderr.

README.md
```diff
@@ -1,19 +1,31 @@
 # data-and-algorithms
 Let's go over data structures and algorithms together! 
-We will use Cracking the Coding Interview as reference.
+We will use [Cracking the Coding Interview Book](http://www.crackingthecodinginterview.com/) as reference.
 
 Feel free to add your solutions in your channel of your choice 😃
 
+Run this script to create the folder structure for your favorite programming language
```
lhchavez: you should use `bash` instead of `sh`. otherwise you lose a lot of functionality!

createFolders.sh
```diff
@@ -0,0 +1,18 @@
...
+PROJECT_NAME=data-and-algorithms
+DIRECTORY=${PWD}
+BASE_DIRECTORY=$(echo "$DIRECTORY" | rev | cut -d "/" -f1 | rev)
+
+if [[("$1" != "")]]; then
+   if [[("$BASE_DIRECTORY" != "$PROJECT_NAME")]]; then
+      echo "Make sure you are in $PROJECT_NAME project"
+      exit
+   fi
+   rootfileName=$1
```
lhchavez: you should always quote all your variables. otherwise if they contain a space, they will be expanded to unexpected stuff!

```shell
bash createFolders.sh "hello world"
```

would cause this line to expand to

```shell
rootfileName=hello world
```

which would try to invoke `world` with the `rootfileName=hello` environment variable.

also, if this is the programming language, maybe call it `programmingLanguage`.

createFolders.sh
```diff
@@ -0,0 +1,18 @@
...
+   mkdir ./$rootfileName
```
lhchavez: here you can do

```shell
mkdir "${DIRECTORY}/${rootfileName}"
```

you should always try to brace the variables to make it super clear where they end. otherwise things like `$myvar_`would confuse readers. is the intention to be a variable called `myvar_` or is it `$myvar` followed by an underscore?

also, by using the `rsync` or `find` alternatives sugegsted below, this `mkdir` is unneeded.

createFolders.sh
```diff
@@ -0,0 +1,18 @@
...
+   cp -a ./Python/ ./$rootfileName/
```
lhchavez: try to use the long forms of flags: https://xkcd.com/1168/

createFolders.sh
```diff
@@ -0,0 +1,18 @@
...
+   rm ./$rootfileName/*/*/*.py
```
lhchavez: maybe it's better to not copy the files in the first place:

```shell
rsync --archive --include='*/' --exclude='*' "${DIRECTORY}/Python/" "${DIRECTORY}/${programmingLanguage}"
```

source: https://stackoverflow.com/a/40586240/688337

or if you want to avoid rsync for whatever reason

```shell
(cd "${DIRECTORY}/Python/" && \
 find . -type d -depth -exec mkdir -p "${DIRECTORY}/${programmingLanguage}/{}" \;)

File: README.md [hunk 8]
```diff
@@ -1,19 +1,31 @@
...
+`sh createFolders.sh  <Programming Language Name>`
```
lhchavez: tiny naming nit: in POSIX, these are called [directories](https://en.wikibooks.org/wiki/A_Quick_Introduction_to_Unix/Directory_Structure), not folders.

darrylbalderas:
Let me add these changes. Gracias amigo 👍

createFolders.sh
```diff
@@ -1,18 +1,12 @@
...
-else
-   echo "Missing folder name: sh createFolders.sh <folder_name>"
-fi
\ No newline at end of file
+if [[ $# -lt 1 ]]; then
+  echo "Usage: bash createFolders.sh <programming language>" >&2
+  exit 1
+fi
+
+programmingLanguage=$1
```
lhchavez: nit: remember the double quotes or you'll get weird results if somebody decides to pass in spaces within an argument. https://google.github.io/styleguide/shell.xml?showone=Quoting#Quoting

Rule of thumb is: every single var should be double quoted AND braced (except for the bash builtin ones, so `$1` and `$#` are fine. https://google.github.io/styleguide/shell.xml?showone=Variable_expansion#Variable_expansion

and in general read that whole style guide, it's great!

createFolders.sh
```diff
@@ -0,0 +1,12 @@
...
+DIRECTORY="$(dirname "$(readlink "$0")")"
+
+if [[ $# -lt 1 ]]; then
+  echo "Usage: bash createFolders.sh <programming language>" >&2
+  exit 1
+fi
+
+programmingLanguage=$1
+mkdir "${DIRECTORY}/${programmingLanguage}"
+rsync --archive --include='*/' --exclude='*' "${DIRECTORY}/Python/" "${DIRECTORY}/${programmingLanguage}"
```
lhchavez: nit: break long lines for better readability.

```suggestion
rsync --archive --include='*/' --exclude='*' "${DIRECTORY}/Python/" \
    "${DIRECTORY}/${programmingLanguage}"
```

createFolders.sh
```diff
@@ -0,0 +1,12 @@
...
+
+DIRECTORY="$(dirname "$(readlink "$0")")"
+
+if [[ $# -lt 1 ]]; then
+  echo "Usage: bash createFolders.sh <programming language>" >&2
+  exit 1
+fi
+
+programmingLanguage=$1
+mkdir "${DIRECTORY}/${programmingLanguage}"
```
lhchavez: you should be able to drop this now since rsync will create it. if it doesn't, it might be because i forgot to add a trailing `/` to the destination directory ^^;

createFolders.sh
darrylbalderas: I'll stay consistent and add the double quotes 💯

createFolders.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+DIRECTORY="$(dirname "$(readlink "$0")")"
```
lhchavez: note that you do need to use the boilerplate i shared as-is to avoid some issues on _some_ POSIX systems:

```suggestion
DIRECTORY="$(readlink --canonicalize -- "$(dirname -- "$0")")"
```

the `--canonicalize` flag removes any symlinks in the path, and the `--` are needed because if some component of the path starts with `-`, the commands _may_ interpret as a flag instead of a path.

createFolders.sh
```diff
@@ -0,0 +1,12 @@
...
+
+if [[ $# -lt 1 ]]; then
+  echo "Usage: bash createFolders.sh <programming language>" >&2
```
lhchavez: WDYT about renaming this file to `createDirectories` for consistency with the POSIX terminology?

darrylbalderas: I think that is fair to have it be consistent with POSIX terminology

createFolders.sh
darrylbalderas: Does `readlink` have a `canonicalize` flag ?  I didn't see that option in terminal manual file for `readlink`

lhchavez: argh. does it have an `-f` flag? i was trying to avoid the shorthand names.

darrylbalderas: screen shot 2019-02-04 at 10 22 31 am

darrylbalderas: I just see the -f option for the stat command

lhchavez: yuck, portable bash programming is hard.

seems like you'll have to live with some platforms being broken. or convert this into Python code.

darrylbalderas: That was going to be my first choice in using python but I was like well let's give bash a try 😉

lhchavez: i just realized there's a better alternative:

```suggestion
DIRECTORY="git rev-parse --show-toplevel
```

this should work **everywhere**.

darrylbalderas: Works like a charm 👍

README.md
```diff
@@ -1,19 +1,31 @@
 # data-and-algorithms
 Let's go over data structures and algorithms together! 
-We will use Cracking the Coding Interview as reference.
+We will use [Cracking the Coding Interview Book](http://www.crackingthecodinginterview.com/) as reference.
 
 Feel free to add your solutions in your channel of your choice 😃
```
lhchavez: now that you're editing this, can you change `channel` for `programming language`?

createDirectories.sh
```diff
@@ -0,0 +1,12 @@
...
+
+DIRECTORY="$(git rev-parse --show-toplevel)"
+
+if [[ $# -lt 1 ]]; then
+  echo "Usage: bash createDirectories.sh <programming language>" >&2
+  exit 1
+fi
+
+programmingLanguage="$1"
+rsync --archive --include='*/' --exclude='*' "${DIRECTORY}/Python/" \
```
lhchavez: oh i just forgot a crucial detail: git never stores empty directories! so you might want to create empty files afterwards:

```bash
find "${DIRECTORY}/${programmingLanguage}" -mindepth 2 -maxdepth 2 -exec touch {}/empty.txt \;
```

darrylbalderas: Oh that's right 😢

lhchavez: i just got merge privileges back! do you want to fix this before it's merged?

darrylbalderas: let me do that

PR #26, opened 2019-02-22
Refactor JavaScript Solutions to 1.2, 1.3, 1.4, 1.6

rroque98:
Implemented changes to JS solutions 1.2, 1.3, 1.4, and 1.6 based on feedback from larger pull request: https://github.com/techqueria/data-and-algorithms/pull/21

### DIFF
--- a/JavaScript/chapter01/1.2 - Check Perm/rroque98_sol.js
+++ b/JavaScript/chapter01/1.2 - Check Perm/rroque98_sol.js
@@ -1,36 +1,34 @@
 /*Given two strings, write a method to decide
 if one is apermutation of the other permutation:*/
 
-const isPermutation = (str1, str2) =>{
+const isPermutation = (str1, str2) => {
   if (str1.length !== str2.length) {
     return false;
   }
-  const obj1 = determineCharCount(str1);
-  const obj2 = determineCharCount(str2);
-  for (let char of str1) {
-    if (obj1[char] !== obj2[char]) {
+  const freq1 = determineCharCount(str1);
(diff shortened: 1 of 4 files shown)


JavaScript/chapter01/1.2 - Check Perm/rroque98_sol.js
```diff
@@ -15,15 +15,14 @@ const isPermutation = (str1, str2) =>{
   return true;
   //******** helper function ******
   function determineCharCount(string) {
-    const obj = {};
-    for (let char of string) {
-      if (obj[char] !== undefined) {
-        obj[char]++;
+    return string.split('').reduce((acc, char) => {
+      if (acc[char] !== undefined) {
```
lhchavez: it might be a bit better to use a [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) like you do in https://github.com/techqueria/data-and-algorithms/pull/26/files#diff-440386f91fbaa9553b4d4652c2d527acR8

Since [`Map.set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/set) supports chaining, you can just do

```javascript
return string
    .split('')
    .reduce((acc, char) => acc.set(char, (acc.get(char) || 0) + 1),
            new Map());
```

rroque98: Thank you for the feedback. I just submitted the changes to use Map instead

PR #27, opened 2019-02-22
Refactor JavaScript 1.5 One Away solution

rroque98:
- Implement changes based on feedback from larger pull request:
https://github.com/techqueria/data-and-algorithms/pull/21

- Attempt to reorganize and simplify logic in order to improve readability
- Add missing test cases

### DIFF
--- a/JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
+++ b/JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
@@ -4,65 +4,65 @@ Given 2 strings write a function that will check if it is
 1 or 0 edits away*/
 
 const isOneAway = (str1, str2) => {
+  const diffInLen = Math.abs(str1.length - str2.length);
+  if (diffInLen > 1) {
+    return false;
+  }
+
+  let errorCount = 0;
   if (str1.length === str2.length) {
-    let errorCount = 0;
     for (let i = 0; i < str1.length; i++) {
-      if (str1[i] === str2[i]) {
-        continue;
-      }
-      errorCount++;
-      if (errorCount > 1) {
-        return false;
+      if (str1[i] !== str2[i]) {
+        errorCount++;
+        if (errorCount > 1) {
+          return false;
+        }
       }
     }
   } else {
-    let errorCount = 0;
-    const longestStr = findLongestStr(str1, str2);
-    let x = 0;
-    for(let i = 0; i < longestStr.length; i++) {
-      let currentChar1 = str1[i];
-      let currentChar2 = str2[x]
-      let nextChar1 = str1[i + 1];
-      let nextChar2 = str2[x + 1];
-      if (str1[i] === str2[x]) {
-        x++;
-        continue;
+    const longStr = str1.length > str2.length ? str1 : str2;
+    const shortStr = str1.length < str2.length ? str1 : str2;
+    for (
+      let i = 0, x = 0;
+      i < longStr.length && x < shortStr.length;
+      i++, x++
+    ) {
+      if (longStr[i] !== shortStr[x]) {
+        errorCount++;
+        if (errorCount > 1) {
+          return false;
+        }
+        i++;
       }
-      errorCount++;
-      if (errorCount > 1) {
-        return false;
-      }
-      if (currentChar1 === nextChar2) {
-        x++;
-        i--;
-        continue;
-      } else if (currentChar2 === nextChar1) {
-        continue;
-      }
-      errorCount++;
-      return false;
-    }
-    // ***** Helper functions ********
-    function findLongestStr(str1, str2) {
-      return (str1.length > str2.length ? str1 : str2);
     }
   }
   return true;
-}
+};
 
 // TESTS
-console.log(isOneAway('pale', 'ple') === true);
-console.log(isOneAway('pales', 'pale') === true);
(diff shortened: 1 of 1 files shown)


JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
```diff
@@ -4,40 +4,65 @@ Given 2 strings write a function that will check if it is
...
 const isOneAway = (str1, str2) => {
+  const diffInLen = Math.abs(str1.length - str2.length);
+  if (diffInLen > 1) {
+    return false;
+  }
+
+  let errorCount = 0;
   if (str1.length === str2.length) {
-    let errorCount = 0;
     for (let i = 0; i < str1.length; i++) {
```
lhchavez: it might be better to use the [guard clause pattern](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html):

```javascript
if (str1.length === str2.length) {
  // ..
  return true;
}
// rest of the code.
```

JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
```diff
@@ -4,40 +4,65 @@ Given 2 strings write a function that will check if it is
...
       }
     }
   } else {
-    let errorCount = 0;
-    const longestStr = findLongestStr(str1, str2);
-    let x = 0;
-    for(let i = 0; i < longestStr; i++) {
-      if (str1[i] !== str2[x]) {
+    const longStr = str1.length > str2.length ? str1 : str2;
+    const shortStr = str1.length < str2.length ? str1 : str2;
```
lhchavez: this is a non-issue in this particular piece of code since you are guaranteed that the lengths of the strings are different, but the opposite of `>` is `<=`, not `<`.

JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
```diff
@@ -4,40 +4,65 @@ Given 2 strings write a function that will check if it is
...
+    for (
+      let i = 0, x = 0;
+      i < longStr.length && x < shortStr.length;
+      i++, x++
+    ) {
+      if (longStr[i] !== shortStr[x]) {
```
lhchavez: more opportunities to use the guard clause pattern!

```javascript
if (longStr[i] === shortStr[x]) {
  continue;
}
```

JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
```diff
@@ -4,40 +4,65 @@ Given 2 strings write a function that will check if it is
...
-    const longestStr = findLongestStr(str1, str2);
-    let x = 0;
-    for(let i = 0; i < longestStr; i++) {
-      if (str1[i] !== str2[x]) {
+    const longStr = str1.length > str2.length ? str1 : str2;
+    const shortStr = str1.length < str2.length ? str1 : str2;
+    for (
+      let i = 0, x = 0;
+      i < longStr.length && x < shortStr.length;
+      i++, x++
```
lhchavez: given that at any point in time `i === x + errorCount`, would it be possible to just use two variables? also, it is customary to use `i`, `j` for indices. you could alternatively use `longIdx` and `shortIdx` if you want to use more descriptive names, but `i`, `j` is okay.

JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
```diff
@@ -4,40 +4,65 @@ Given 2 strings write a function that will check if it is
...
   return true;
-}
+};
 
 // TESTS
-console.log(isOneAway('pale', 'ple') === true);
-console.log(isOneAway('pales', 'pale') === true);
-console.log(isOneAway('pale', 'bale') === true);
-console.log(isOneAway('pale', 'bake') === false);
+console.log(isOneAway('pale', 'ple') === true); // deletion
```
lhchavez: maybe try restructuring the tests like this:

```javascript
function runTests(cases, expected) {
  for (const [str1, str2] of cases) {
    console.log(isOneAway(str1, str2) === expected);
  }
}

runTests([
  ['pale', 'opale'], // insertion in beginning
  // ...
], true);
```

JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
rroque98: Yes, TY. I am working on being more aware of this pattern

JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
rroque98: Thanks for the feedback. Definitely helps make the code more DRY

JavaScript/chapter01/1.5 - OneAway/rroque98_sol.js
rroque98: Good catch. There was no need for the variable `x` which also helped reduce the amount of code in this for loop. Also, in regards to the indices, sometimes, I'm not a huge fan of `i` and `j` since they look so similar, but yes, I agree. I'm going to try to use `i` and `j` instead since this is more customary.

rroque98:
This looks like it has been approved and just needs to be merged. Thanks again for the very helpful feedback.

PR #29, opened 2019-02-22
Add JavaScript solution to 1.8 Zero Matrix

rroque98:
Implement feedback from larger pull request: https://github.com/techqueria/data-and-algorithms/pull/21

### DIFF
--- a/JavaScript/chapter01/1.8 - Zero Matrix/rroque98_sol.js
+++ b/JavaScript/chapter01/1.8 - Zero Matrix/rroque98_sol.js
@@ -6,50 +6,71 @@ const zeroMatrix = nestedArr => {
   const rowsAndColsWithZeros = checkForZeroIndex(nestedArr);
   const rowLength = nestedArr[0].length;
   const columnLength = nestedArr.length;
-  for (const rowIndex of rowsAndColsWithZeros['rows']) {
-    for (let x = 0; x < rowLength; x++) {
-      nestedArr[rowIndex][x] = 0;
+  for (const row in rowsAndColsWithZeros['rows']) {
+    for (let col = 0; col < rowLength; col++) {
+      nestedArr[row][col] = 0;
     }
   }
-  for (const colIndex of rowsAndColsWithZeros['columns']) {
-    for (let x = 0; x < columnLength; x++) {
-      nestedArr[x][colIndex] = 0;
+  for (const col in rowsAndColsWithZeros['columns']) {
+    for (let row = 0; row < columnLength; row++) {
+      nestedArr[row][col] = 0;
     }
   }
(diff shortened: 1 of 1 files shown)


JavaScript/chapter01/1.8 - Zero Matrix/rroque98_sol.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+  for (const colIndex of rowsAndColsWithZeros['columns']) {
+    for (let x = 0; x < columnLength; x++) {
+      nestedArr[x][colIndex] = 0;
+    }
+  }
+  return nestedArr;
+
+  // **** Helper functions ****
+  function checkForZeroIndex(nestArr) {
```
lhchavez: If you make this a free function instead of nesting it within this other function, you can test it independently.

JavaScript/chapter01/1.8 - Zero Matrix/rroque98_sol.js
```diff
@@ -0,0 +1,55 @@
...
+    let zeroIndices = { rows: [], columns: [] };
+    for (let i = 0; i < nestArr.length; i++) {
+      for (let x = 0; x < nestArr[i].length; x++) {
+        if (nestArr[i][x] === 0) {
+          zeroIndices['rows'].push(i);
```
lhchavez: what would the runtime complexity of a matrix full of zeroes be?

rroque98: If we define N as the number of rows and M as the number of columns in the matrix, then I believe the time complexity would be O(NM)

JavaScript/chapter01/1.8 - Zero Matrix/rroque98_sol.js
rroque98: I made the helper function free now, and I included some tests for the helper function

JavaScript/chapter01/1.8 - Zero Matrix/rroque98_sol.js
lhchavez: it is after the update, but it used to be O(max(N²M, NM²)), since the `zeroIndices` matrix would have up to O(max(N, M)) zeroes for each row and column, and each one of those would cause up to O(max(N, M)) updates per row/column.

now you're just counting each row/column just once.

rroque98: I see. I noticed I was double counting rows and columns that were already in `zeroIndices` after I made the helper function free and tried testing it for the desired behavior. Thanks for catching this and for the explanation.

PR #30, opened 2019-02-22
updates

camilosalazar98:
Finished 3.2 but I need to make a unit test for every question


fvntr:
@lhchavez we can add a prettier file for spacing

PR #31, opened 2019-02-25
Solutions to problems 1.3 - 1.6

lomeli-noe:


Python/chapter01/1.3 - URLify/lomeli_1.3.py
```diff
@@ -0,0 +1,14 @@
...
+    temp = []
+    url = '%20'
+    for i in range(size):
+        if string[i] == ' ':
+            temp.append(url)
+        else:
+            temp.append(string[i])
+    return ''.join(temp)
+
+print(Urlify("Mr John Smith     ", 13))
```
lhchavez: It might be better to use the [unittest](https://docs.python.org/3/library/unittest.html) library since that will let you know when a test fails instead of you having to manually go through the output.

Python/chapter01/1.3 - URLify/lomeli_1.3.py
```diff
@@ -0,0 +1,14 @@
+# Write a method to replace all spaces in a string with '%20'
+# "Mr John Smith     ", 13 => "Mr%20John%20Smith"
+
+def Urlify(string, size):
+    temp = []
+    url = '%20'
```
lhchavez: nit: maybe call this something a bit better, like `encoded_space`, since this is not an URL.

Python/chapter01/1.3 - URLify/lomeli_1.3.py
```diff
@@ -0,0 +1,14 @@
+# Write a method to replace all spaces in a string with '%20'
+# "Mr John Smith     ", 13 => "Mr%20John%20Smith"
+
+def Urlify(string, size):
+    temp = []
```
lhchavez: nit: it might be better to give this a more [descriptive name](https://medium.com/coding-skills/clean-code-101-meaningful-names-and-functions-bf450456d90c) than `temp`. something like `urlified` could work.

Python/chapter01/1.3 - URLify/lomeli_1.3.py
```diff
@@ -0,0 +1,14 @@
...
+    url = '%20'
+    for i in range(size):
```
lhchavez: it might be more [Pythonic](https://docs.python-guide.org/writing/style/) to use

```python
for c in string[:size]:
```

Python/chapter01/1.4 - PalinPerm/lomeli_1.4.py
```diff
@@ -0,0 +1,24 @@
...
+
+def PalinPerm(str):
+    # Make all letters lowercase
+    str = str.lower()
+    # Remove spaces from string
+    str = ''.join(str.split())
+    uniq = set(c for c in str)
+
+    if len(str) % 2 != 0:
+        return (len(str) / 2)+1 == len(uniq)
```
lhchavez: try using [yapf](https://github.com/google/yapf) to format your code so that the spacing between operators is consistent.

Python/chapter01/1.4 - PalinPerm/lomeli_1.4.py
```diff
@@ -0,0 +1,24 @@
...
+    else:
+        return (len(str) / 2) == len(uniq)
+
+print(PalinPerm("Tact toa"))
+print(PalinPerm("reca Cra"))
+print(PalinPerm("Taco dog"))
+print(PalinPerm("ab Ba"))
+print(PalinPerm("abcdba"))
```
lhchavez: what would the answer to `PalinPerm("abaa")` be?

Python/chapter01/1.5 - OneAway/lomeli_1.5.py
```diff
@@ -0,0 +1,28 @@
...
+
+    for c in smallStr:
+        bigStr = bigStr.replace(c,'')
+
+    return len(bigStr) <= 1
+
+print(OneAway("pale", "ple"))
+print(OneAway("pales", "pale"))
+print(OneAway("pale", "bale"))
+print(OneAway("pale", "bake"))
```
lhchavez: what should the answer to `OneAway("aaaaaaaaaaaaaaaaaaaa", "a")` be?

Python/chapter01/1.6 - String Compression/lomeli_1.6.py
```diff
@@ -0,0 +1,29 @@
+# Implement a method to perform basic string compression using the counts
+# of repeated characters.  If the "compressed" string would not become smaller
+# than the original string, your method should return the original stringself.
+# You can assume that the string has only uppercase and lowercase letters.
+
+def stringCompress(st):
+    result = []
+    cur = st[0]
+    count = 0
+    for i in range(len(st)):
```
lhchavez: given that you don't use the index (by itself) for anything, it's better to do:

```python
for c in st:
```

if you find yourself needing both the index and the character, you can use

```python
for i, c in enumerate(st):
```

Python/chapter01/1.6 - String Compression/lomeli_1.6.py
```diff
@@ -0,0 +1,29 @@
...
+    comp = ''.join(result)
+    if len(comp) < len(st):
+        return comp
+    else:
+        return st
+
+print(stringCompress("aabcccccaaa"))
+print(stringCompress("aabca"))
+print(stringCompress("XxxyyyZZzzzzz"))
+print(stringCompress("abc"))
```
lhchavez: what should the answer to

```python
stringCompress("")
```

be?

Python/chapter01/1.6 - String Compression/lomeli_1.6.py
```diff
@@ -0,0 +1,29 @@
...
+# of repeated characters.  If the "compressed" string would not become smaller
+# than the original string, your method should return the original stringself.
+# You can assume that the string has only uppercase and lowercase letters.
+
+def stringCompress(st):
+    result = []
+    cur = st[0]
+    count = 0
+    for i in range(len(st)):
+        if cur == st[i]:
```
lhchavez: you can use the [guard clause pattern](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html):

```python
if cur == c:
  count += 1
  continue
# rest of the code
result.append(cur)
...
```

lhchavez (commented):
a bit about code review etiquette: normally you want to keep [all your changes small](https://smartbear.com/learn/code-review/best-practices-for-peer-code-review/). by splitting your change in smaller chunks you'll be able to make progress faster since non-controversial chunks will be approved and merged faster. since each individual program is independent of each other, it is ideal to make each one be its own review. you'll get better mileage out of the small network of (volunteer!) reviewers.

and another tip: try running a code formatter like [yapf](https://github.com/google/yapf) so that it has consistent use of space. that way the code review can focus on actual content and less on presentation. try also running [flake8](https://gitlab.com/pycqa/flake8) on your code so that it catches some common style mistakes.

PR #33, opened 2019-03-31
Miguel 1.6 - String Compression [Python]

miguelHx:


Python/chapter01/1.6 - String Compression/miguel_1.6_sol.py
```diff
@@ -0,0 +1,61 @@
...
+            continue
+        compressed.append(prev_char)
+        compressed.append(str(count)[0])
+        prev_char = c
+        count = 1
+    # clean up: last character count
+    compressed.append(prev_char)
+    compressed.append(str(count)[0])
+
+    return ''.join(compressed) if len(compressed) < len(s) else s
```
lhchavez: in the spirit of the guard clause pattern, you can do

```python
if len(compressed) > len(s):
  return s
return ''.join(compressed)
```

in general try to avoid cramming too much logic in simple statements, and only use the ternary syntax where you absolutely need an expression (like within a comprehension).

Python/chapter01/1.6 - String Compression/miguel_1.6_sol.py
```diff
@@ -0,0 +1,61 @@
...
+            ('aabbccdd', 'aabbccdd'),
+            ('', ''),
+            ('c', 'c'),
+            ('aaAAccCCCCC', 'a2A2c2C5')
+        ]
+        for s, expected in cases:
+            self.assertEqual(f(s), expected, msg=(s, expected))
+
+    def test_string_compression(self):
+        self._run_tests(str_compression)
```
lhchavez: since you are only invoking this function in one place and leaving it as standalone doesn't improve readability, you can inline `_run_tests` here (to improve readability).

Python/chapter01/1.6 - String Compression/miguel_1.6_sol.py
```diff
@@ -0,0 +1,61 @@
...
+    Space Complexity: O(n), compressed list
+    :param s: the string we want to compress
+    :return: the compressed string or s if the compressed string is not smaller
+    """
+    if len(s) == 0:
+        return s
+    compressed = []
+    prev_char = s[0]
+    count = 1
+    for c in s[1:len(s)]:
```
lhchavez: nit: you can omit the upper bound if it's the length of the sequence:

```python
for c in s[1:]:
```

Python/chapter01/1.6 - String Compression/miguel_1.6_sol.py
```diff
@@ -0,0 +1,61 @@
...
+    compressed.append(prev_char)
+    compressed.append(str(count)[0])
+
+    return ''.join(compressed) if len(compressed) < len(s) else s
+
+
+class TestStringCompressionFunction(unittest.TestCase):
+    def _run_tests(self, f: Callable[[str], str]) -> None:
+        cases = [
+            ('aabcccccaaa', 'a2b1c5a3'),
```
lhchavez: what happens with

```python
    ('aaaaaaaaaa', 'a10'),
```

?

Python/chapter01/1.6 - String Compression/miguel_1.6_sol.py
```diff
@@ -0,0 +1,61 @@
...
+    if len(s) == 0:
+        return s
+    compressed = []
+    prev_char = s[0]
+    count = 1
+    for c in s[1:len(s)]:
+        if c == prev_char:
+            count += 1
+            continue
+        compressed.append(prev_char)
```
lhchavez: can you use [`str.format`](https://docs.python.org/3/library/stdtypes.html#str.format) to coalesce both appends and make it a bit more readable?

miguelHx:
updated. what is your view on f strings? https://realpython.com/python-f-strings/

Python/chapter01/1.6 - String Compression/miguel_1.6_sol.py
```diff
@@ -0,0 +1,59 @@
...
+    if len(s) == 0:
+        return s
+    compressed = []
+    prev_char = s[0]
+    count = 1
+    for c in s[1:]:
+        if c == prev_char:
+            count += 1
+            continue
+        compressed.append('{}{}'.format(prev_char, str(count)))
```
lhchavez: final nit: you don't need to perform manual stringification on `count`, since the `format` method can do that for you.

PR #34, opened 2019-04-21
Miguel 1.7 - Rotate Matrix [Python]

miguelHx:


miguelHx:
stand by, I need to remove the commits before d86040c lol

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
```diff
@@ -0,0 +1,174 @@
...
+    :return: a newly rotated matrix
+    """
+    rotated = [[0 for i in range(N)] for j in range(N)]
+    for i in range(N):
+        for j in range(N):
+            rotated[j][(N-1)-i] = matrix[i][j]
+    return rotated
+
+
+def rotate_matrix_in_place(matrix: List[List[int]], start_row: int, start_col: int, N: int) -> List[List[int]]:
```
lhchavez: you can provide defaults to all parameters except `matrix`:

```python
def rotate_matrix_in_place(
    matrix: List[List[int]],
    start_row: int = 0,
    start_col: int = 0,
    N: Optional[int]) -> List[List[int]]:
  """
  ...
  """
  if N is None:
    N = len(matrix)
```

that way you can invoke both methods exactly the same!

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
```diff
@@ -0,0 +1,174 @@
...
+Python version 3.7.0
+1.7 - Rotate Matrix
+Given an image represented by an NxN matrix, where each pixel in the image is 4 bytes, write a method to rotate
+the image by 90 degrees.  Can you do this in place?
+"""
+import unittest
+from typing import List
+
+
+def rotate_matrix(matrix: List[List[int]], N: int) -> List[List[int]]:
```
lhchavez: nit: you can drop the `N` parameter since it's equal to `len(matrix)`.

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
```diff
@@ -0,0 +1,174 @@
...
+            rotated[j][(N-1)-i] = matrix[i][j]
+    return rotated
+
+
+def rotate_matrix_in_place(matrix: List[List[int]], start_row: int, start_col: int, N: int) -> List[List[int]]:
+    """
+    Does the same as rotate_matrix, but in place.
+    Runtime: O(N^2), asymptotic runtime depends on N. We make N^2 swaps.
+    Space Complexity: O(1), constant amount of temp variables that does not depend on N.
+                        Note: I am assuming that compiler will optimize the tail recursion.
```
lhchavez: this assumption is unfortunately false and will most likely never be true: http://neopythonic.blogspot.com/2009/04/tail-recursion-elimination.html

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
```diff
@@ -0,0 +1,174 @@
...
+        for r in range(0, num_rotations):
+            temp = temp_new
+            # compute new rotated indices
+            prev_col = rotated_col
+            rotated_col = N - 1 - rotated_row + (start_row * 2)  # offset to account for reduced N
+            rotated_row = prev_col
+            # store value at newly computed indices
+            temp_new = matrix[rotated_row][rotated_col]
+            matrix[rotated_row][rotated_col] = temp
+        if col - start_col >= N - 2:
```
lhchavez: do you really need the `while True`? can you try to model this as a regular for loop?

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
```diff
@@ -0,0 +1,174 @@
...
+            break
+        col = col + 1
+    return rotate_matrix_in_place(matrix, start_row + 1, start_col + 1, N - 2)
```
lhchavez: same here: you are mixing two different styles of looping, and both can be easily replaced with a regular for loop.

one thing you _can_ do is extract part of the functionality of this function to a helper function, if you find that to be cleaner.

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
```diff
@@ -0,0 +1,174 @@
...
+    def setUp(self):
+        self.cases = [
+            (
+                [
+                    [1, 2, 3, 4],
+                    [5, 6, 7, 8],
+                    [9, 10, 11, 12],
+                    [13, 14, 15, 16]
+                ],
+                4,
```
lhchavez: with the changes suggested above, you don't need this.

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
miguelHx: Does that mean that the space complexity will not be O(1)?

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
miguelHx: The reason I used the while True loop was to emulate a do-while loop due to the for loop not entering when N was 2 or 3.  Will try to model as a regular for loop though.

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
miguelHx: Dropped

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
miguelHx: Extracted the rotations to a helper function.  Does it follow convention to prepend this new helper function with '_' since it is used internally?

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
lhchavez: correct: Python will always use O(n) stack space.

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
```diff
@@ -0,0 +1,185 @@
...
+    """
+    N = len(matrix)
+    rotated = [[0 for i in range(N)] for j in range(N)]
+    for i in range(N):
+        for j in range(N):
+            rotated[j][(N-1)-i] = matrix[i][j]
+    return rotated
+
+
+def _perform_full_rotation(matrix: List[List[int]], row: int, col: int, N: Optional[int] = None):
```
lhchavez: since this is a helper function, you can always call it with the correct parameters: the `Optional[int]` is not needed here.

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
```diff
@@ -0,0 +1,185 @@
...
+    """
+    Helper function that performs four 90 degree rotations starting at row, col.
+    :param matrix: an NxN matrix
```
lhchavez: note: the matrix is not an `NxN` matrix, since this is only processing an `NxN` submatrix out of an `MxM` matrix (where `M = len(matrix)`).

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
```diff
@@ -0,0 +1,185 @@
...
+    :param row: starting row
+    :param col: starting column
+    :param N: size of matrix
```
lhchavez: nit: size of **sub**matrix.

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
```diff
@@ -0,0 +1,185 @@
...
+    :param matrix: an NxN matrix
+    :param start_row: starting row index
+    :param start_col: starting col index
+    :param N: the size of the matrix (NxN)
+    :return: the input matrix, but rotated
+    """
+
+    if N is None:
+        N = len(matrix)
+    if N == 0 or N == 1:
```
lhchavez: what you are doing in this method is initializing:

```
start_row = 0
start_col = 0 # this is identical to start_row, so you don't really need it
N = len(matrix)
```

then repeatedly do:

```python
_perform_full_rotation(matrix, start_row, start_col, N)
start_row += 1
start_col += 1
N -= 2
```

until `N < 2`. That looks and feels like a for loop. furthermore, you can derive those three variables from the loop variable.

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
```diff
@@ -0,0 +1,185 @@
...
+        N = len(matrix)
+    num_rotations = 4
+    start_row = row
+    rotated_row = row
+    rotated_col = col
+    temp_new = matrix[row][col]
+    for r in range(0, num_rotations):
+        temp = temp_new
+        # compute new rotated indices
+        prev_col = rotated_col
```
lhchavez: note that in Python you can replace

```python
tmp = a
a = b
b = tmp
```

with

```python
a, b = b, a
```

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
lhchavez: it does!

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
miguelHx: The for loop you are suggesting would get rid of the recursive calls, right?  Do you recommend doing it that way instead of the recursive approach?  If so, why?

lhchavez: that's correct. loops are significantly easier to understand than recursion (in most cases). the fact that the loop uses O(1) memory is an added bonus.

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    Does the same as rotate_matrix, but in place.
+    Runtime: O(N^2), asymptotic runtime depends on N. We make N^2 swaps.
+    Space Complexity: O(1), constant amount of temp variables that does not depend on N.
+    :param matrix: an NxN matrix
+    :return: the input matrix, but rotated
+    """
+    start_row = 0
+    N = len(matrix)
+    for n in range(N, 1, -2):
```
lhchavez: nit:

```python
for start_row, n in enumerate(range(N, 1, -2)):
```

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
```diff
@@ -0,0 +1,172 @@
...
+    num_rotations = 4
+    start_row = row
+    rotated_row = row
+    rotated_col = col
+    temp_new = matrix[row][col]
+    for r in range(0, num_rotations):
+        # compute new rotated indices
+        prev_col = rotated_col
+        rotated_col = M - 1 - rotated_row + (start_row * 2)  # offset to account for reduced N
+        rotated_row = prev_col
```
lhchavez: nit:

```python
rotated_col, rotated_row = M - 1 - rotated_row + (start_row * 2), rotated_col
```

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
```diff
@@ -0,0 +1,172 @@
...
+    :param col: starting column
+    :param M: size of sub-matrix
+    :return:
+    """
+    num_rotations = 4
+    start_row = row
+    rotated_row = row
+    rotated_col = col
+    temp_new = matrix[row][col]
+    for r in range(0, num_rotations):
```
lhchavez: nit: you are not using `r`, so you can rename this as `_`.

Python/chapter01/1.7 - Rotate Matrix/miguel_1.7_sol.py
```diff
@@ -0,0 +1,172 @@
...
+    Runtime: O(N^2), asymptotic runtime depends on N. We make N^2 swaps.
+    Space Complexity: O(1), constant amount of temp variables that does not depend on N.
+    :param matrix: an NxN matrix
+    :return: the input matrix, but rotated
+    """
+    start_row = 0
+    N = len(matrix)
+    for n in range(N, 1, -2):
+        for col in range(start_row, n - 1 + start_row):
+            _perform_full_rotation(matrix, start_row, col, n)
```
lhchavez: you might want to add some comments that overall explain how this method works since it's non-trivial and its workings are split across methods. this increases local readability at the expense of global understandability, but that can be offset with comments so it's the right trade-off to do.

you can mention that every iteration of the outer loop will rotate one of the "rings" of the matrix, the inner loop iterates rotating 4 entries of the ring every time, and you can describe how you rotate the 4 entries either here or in a comment in `_perform_full_rotation`. bonus points for some neat [ASCII diagram](https://blog.regehr.org/archives/1653), but you don't need to go that far if you can convey the idea in prose.

you can see an example of how comments can be used to explain a gnarly piece of code (plus an instance of me judiciously breaking the rule of "prefer loops over recursion"!) here: https://android.googlesource.com/platform/external/minijail/+/255a8e2879edb49f395f819e2f8590b40a8284ed/tools/compiler.py#158

miguelHx: Will add comments, thank you for all the feedback!

miguelHx: Added comments, sorry for the delay, I had exams and many assignments to complete last week.

lhchavez:
it seems like i lost write access to this repository D: @fvntr will need to help us merge it.

PR #35, opened 2019-06-01
Miguel 1.8 - Zero Matrix [Python]

miguelHx:


Python/chapter01/1.8 - Zero Matrix/miguel_1.8_sol.py
```diff
@@ -0,0 +1,135 @@
...
+    :param matrix: an MxN matrix.  M is the number of rows, N is the number of columns
+    :return: a zero-modified matrix
+    """
+    M = len(matrix)
+    N = len(matrix[0])
+
+    zeroed_rows = set()
+    zeroed_cols = set()
+
+    for row in range(M):
```
lhchavez: you can also write this as

```suggestion
    for i, row in enumerate(M):
        for j, num in enumerate(row):
```

which is more internally consistent.

Python/chapter01/1.8 - Zero Matrix/miguel_1.8_sol.py
```diff
@@ -0,0 +1,135 @@
...
+import unittest
+from typing import List
+
+
+def zero_matrix(matrix: List[List[int]]) -> List[List[int]]:
+    """
+    zero_matrix will take in an MxN matrix and when a 0 element is found, the whole column and row
+    will be set to 0.
+    Runtime: O(M * N)
+    Space Complexity: O(min(M, N))
```
lhchavez: can you elaborate why this is `min(M, N)`?

Python/chapter01/1.8 - Zero Matrix/miguel_1.8_sol.py
```diff
@@ -0,0 +1,135 @@
...
+                continue
+            # otherwise, set row to 0 by looping through columns of current row
+            for i in range(N):
+                matrix[row][i] = 0
+            # set column to 0 by looping through rows of current column
+            for j in range(M):
+                matrix[j][col] = 0
+            # update zeroed row and col sets
+            zeroed_cols.add(col)
+            zeroed_rows.add(row)
```
lhchavez: just a small optimization: you could add a `break` here to stop considering the rest of the elements of the row since you know that `row in zeroed_rows` will be `True` in the rest of the elements.

Python/chapter01/1.8 - Zero Matrix/miguel_1.8_sol.py
miguelHx: Initially I thought the worst case for storing zeroed columns and rows was when there were 0s along the diagonal (The diagonal count of an MxN matrix is the smaller value between M and N, hence min(M, N)).  However, I was wrong.
I overlooked the case when all values along a row or column are 0.
Solution and test cases are updated accordingly to account for that.  My new solution takes a different approach.

Python/chapter01/1.8 - Zero Matrix/miguel_1.8_sol.py
```diff
@@ -0,0 +1,179 @@
...
+                    [1, 2, 3, 4],
+                    [5, 6, 7, 8],
+                    [9, 10, 11, 12]
+                ],
+                [
+                    [0, 0, 0, 0],
+                    [0, 0, 0, 0],
+                    [0, 0, 0, 0],
+                    [0, 0, 0, 0]
+                ]
```
lhchavez: can you add

```python
[
    [[0] * 1000 for _ in range(1000)],
    [[0] * 1000 for _ in range(1000)],
],
```

? a O(MN) algorithm should have no trouble with that.

miguelHx: After adding that test case, the code ran and passed the test cases in about 112 seconds.  Very slow.  In the worst case (all 0s in input matrix), would the runtime be O(M*N * (M + N)) = O(M^2*N + N^2*M)?  Because in that case of all 0s input, each element in the matrix will do M + N work.

My solution is very slow. I'm going to try to optimize it.

miguelHx: Updated

Python/chapter01/1.8 - Zero Matrix/miguel_1.8_sol.py
```diff
@@ -0,0 +1,186 @@
+"""
+Python version 3.7.0
+1.8 - Zero Matrix
+Write an algorithm such that if an element in an MxN matrix is 0, its entire row and column are set to 0
```
lhchavez: nit: wordwrap this comment and the one in L12.

miguelHx: How exactly should I word-wrap the comments?

miguelHx: Nvm I think I got it, lmk if the comments are better now

lhchavez: i use vim, where the [`gqip`](https://vim.fandom.com/wiki/Automatic_word_wrapping) command does the trick. all other editors have similar ways of doing word-wrapping for you.

but yes, this looks much better!

PR #36, opened 2019-06-05
Miguel 1.9 - String Rotation [Python]

miguelHx:


Python/chapter01/1.9 - String Rotation/miguel_1.9_sol.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+    # not rotation if lengths don't match
+    if len(s1) != len(s2):
+        return False
+    # if they match, but both empty, return true
+    if s1 == '' and s2 == '':
+        return True
+    size = len(s1)
+    start_idx = -1
+    for i, c in enumerate(s2):
```
lhchavez: you are circumventing the restriction of doing a single call to `is_substring` by kinda sorta checking for a partial substring by hand, which goes against the spirit of the problem.

Python/chapter01/1.9 - String Rotation/miguel_1.9_sol.py
```diff
@@ -25,15 +25,9 @@ def string_rotation(s1: str, s2: str) -> bool:
...
-    compare the rest of the original string to the rotated
-    string's substring of the same length.
-    If they match, then we save the index of the
-    rotated string + 1, and if whatever is left of
-    s1 is a substring of s2, then we must have a
-    rotation since we compared the complementary part already.
-    Runtime: Worst case:  O(n^2)
+    If s2 is a rotation of s1, then when we add s1 to itself, s2
+    must be a substring of s1 + s1
+    Runtime: O(n)
```
lhchavez: can you back this claim?

lhchavez: sorry, should have been more clear: i don't think O(n) is accurate. can you explain how you arrived at that?

Python/chapter01/1.9 - String Rotation/miguel_1.9_sol.py
```diff
@@ -0,0 +1,83 @@
...
+    Choosing an index where we rotate to be any index,
+    let s1 and s2 consist of two parts, a and b.
+    The original string will be s2 = ab.
+    The rotated string will be s1 = ba.
+    Since we are using substring, we want s2 = ab to appear in
+    a larger string ..ab..
+    This is guaranteed to happen if we have s1 + s1 = baba
+    because ab shows up in there.
+    Runtime: worst case: O(n^2) - substring found at end of full string
+    Space Complexity:  O(1)
```
lhchavez: argh, i missed this one. you are now constructing a brand new string, so this is no longer O(1) (although with some clever data structures like Cords it could have been).

miguelHx: Going to update the comment now.

PR #37, opened 2019-06-09
Miguel 2.1 - Remove Dups [Python]

miguelHx:


Python/chapter02/2.1 - Remove Dups/miguel_2.1_sol.py
```diff
@@ -0,0 +1,165 @@
...
+    :param ll: a linked list
+    :return: a linked list without duplicates
+    """
+    n = ll.head
+    unique_vals = set()
+    while n.next is not None:
+        unique_vals.add(n.data)
+        n = n.next
+    unique_vals.add(n.data)
+    return build_linked_list(list(unique_vals))
```
lhchavez: what is the complexity of this operation?

Python/chapter02/2.1 - Remove Dups/miguel_2.1_sol.py
```diff
@@ -0,0 +1,165 @@
...
+    remove_dups will remove duplicates from the
+    input linked list ll. A temporary buffer
+    is used.
+    Runtime: O(N)
+    Space Complexity: O(N)
+    :param ll: a linked list
+    :return: a linked list without duplicates
+    """
+    n = ll.head
+    unique_vals = set()
```
lhchavez: maybe

```suggestion
    unique_vals = set([n.data])
```

to avoid L86

Python/chapter02/2.1 - Remove Dups/miguel_2.1_sol.py
```diff
@@ -0,0 +1,165 @@
...
+            (
+                build_linked_list([1, 1]),
+                build_linked_list([1])
+            ),
+            (
+                build_linked_list([2, 2, 2, 2, 2, 2]),
+                build_linked_list([2])
+            ),
+            (
+                build_linked_list([1, 1, 3, 4, 5, 5, 6, 7]),
```
lhchavez: can you add a couple of test cases where the linked list is not sorted?

Python/chapter02/2.1 - Remove Dups/miguel_2.1_sol.py
```diff
@@ -0,0 +1,165 @@
...
+        while m is not None:
+            if m.data == curr_data:
+                # re-arrange pointers to omit
+                # the duplicate
+                prev.next = m.next
+                m = prev.next
+            else:
+                # otherwise, advance m and prev pointers
+                prev = m
+                m = m.next
```
lhchavez: in both cases, you assign `m = prev.next`. it may be easier to understand if you spell it that way in both branches, and since it's the same in both, might as well make it unconditional.

Python/chapter02/2.1 - Remove Dups/miguel_2.1_sol.py
miguelHx: runtime for list(unique_vals) is O(N)
then to build a linked list from that:  O(N)

Python/chapter02/2.1 - Remove Dups/miguel_2.1_sol.py
miguelHx: Pycharm was complaining to use a set literal instead,
so I just did:
```python
unique_vals = {n.data}
```

Python/chapter02/2.1 - Remove Dups/miguel_2.1_sol.py
lhchavez: can you prove the latter?

miguelHx: Whoops, I overlooked the append_to_tail runtime, which is O(N).  So, this means that to build a linked list from 'numbers' using append_to_tail will be O(N^2).

I'm thinking I can loop from the end of numbers and append to head, which will be O(1) insertion.

Or, I can maintain a 'tail' pointer in the linked list to ensure O(1) insertion.

lhchavez: either work, although the latter might be a bit more pythonic and simpler to understand.

Python/chapter02/2.1 - Remove Dups/miguel_2.1_sol.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+class Node:
+    def __init__(self, d: int):
+        self.data = d
+        self.next = None
+
+
+class LinkedList:
+    def __init__(self, initial_value: int = None):
```
lhchavez:
```suggestion
    def __init__(self, initial_value: Optional[int] = None):
```

Python/chapter02/2.1 - Remove Dups/miguel_2.1_sol.py
```diff
@@ -0,0 +1,181 @@
...
+            return '<empty>'
+        ll = []
+        n = self.head
+        while n.next is not None:
+            ll.append('{} -> '.format(n.data))
+            n = n.next
+        ll.append(str(n.data))
+        return ''.join(ll)
+
+    def __eq__(self, other):
```
lhchavez: nit: can you add typing hints?

Python/chapter02/2.1 - Remove Dups/miguel_2.1_sol.py
```diff
@@ -0,0 +1,181 @@
...
+        while a is not None and b is not None:
+            if a.data != b.data:
+                return False
+            # otherwise, advance both pointers
+            a = a.next
+            b = b.next
+        return a is None and b is None
+
+
+def build_linked_list(numbers: List[int]) -> LinkedList:
```
lhchavez: you can merge this into `LinkedList`'s constructor:

```python
def __init__(self, *numbers: int):
```

that should also take care of the missing `Optional`.

Python/chapter02/2.1 - Remove Dups/miguel_2.1_sol.py
miguelHx: Seems like adding type hints for __eq__ would go something like this:
https://stackoverflow.com/questions/37557411/why-does-defining-the-argument-types-for-eq-throw-a-mypy-type-error

Python/chapter02/2.1 - Remove Dups/miguel_2.1_sol.py
```diff
@@ -0,0 +1,179 @@
...
+        n = n.next
+    return ll
+
+
+class TestRemoveDups(unittest.TestCase):
+
+    def setUp(self):
+        self.test_cases = [
+            (
+                build_linked_list([1, 2, 3, 3, 5]),
```
lhchavez: you can change all of these to use the constructor to be able to remove `build_linked_list` altogether.

```suggestion
                LinkedList(1, 2, 3, 3, 5),
```

PR #38, opened 2019-06-14
Miguel 2.2 - Kth To Last [Python]

miguelHx:


Python/chapter02/2.2 - Return Kth to Last/miguel_2.2_soln.py
```diff
@@ -0,0 +1,163 @@
...
+    Runtime: O(N)
+    Space Complexity: O(N)
+    :param ll: a linked list
+    :param k: an integer where k >= 0
+    :return:
+        kth to last Node from the linked list
+        or None
+    """
+    if k <= 0:
+        return None
```
lhchavez: maybe

```suggestion
        raise IndexError('list index out of range')
```

otherwise the type is wrong: you are promising that you are going to return `Node`, but you are actually returning `Optional[Node]`.

Python/chapter02/2.2 - Return Kth to Last/miguel_2.2_soln.py
```diff
@@ -0,0 +1,163 @@
...
+                LinkedList(1, 2, 3, 4, 5, 6),
+                5,
+                Node(2)
+            ),
+            (
+                LinkedList(1, 2, 3, 4, 5, 6),
+                6,
+                Node(1)
+            ),
+            (
```
lhchavez: Can you move the negative cases to a different test case? that way you can re-express the test as

```python
l = list(range(1, 6))
ll = LinkedList(*l)
for k in range(len(l)):
   this.assertEqual(kth_to_last(ll, k), l[-k-1])
```

Python/chapter02/2.2 - Return Kth to Last/miguel_2.2_soln.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def kth_to_last(ll: LinkedList, k: int) -> Node:
+    """
+    kth_to_last will return the kth to last node
+    from the input linked list.
+    Going to reverse the linked list and then
+    count k steps.
+    Runtime: O(N)
+    Space Complexity: O(N)
```
lhchavez: is there a O(1) answer? is there an answer where you don't explicitly allocate any memory and instead let the runtime handle that? (you might think that is cheating because it would still be O(n) space complexity, but it "feels" O(1)).

miguelHx: I'm thinking that I can loop through the linked list to get the size, then traverse to the size - kth node. (starting i at 1).  No extra linked list created.

Maybe it would be better to keep track of the linked list's size as a member variable?

lhchavez: > I'm thinking that I can loop through the linked list to get the size, then traverse to the size - kth node. (starting i at 1). No extra linked list created.

that works.

> Maybe it would be better to keep track of the linked list's size as a member variable?

that's definitely an option. C++ "recently" changed the standard to actually specify that: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2009/n2923.pdf

Python/chapter02/2.2 - Return Kth to Last/miguel_2.2_soln.py
```diff
@@ -84,79 +87,39 @@ def kth_to_last(ll: LinkedList, k: int) -> Node:
...
-        n = n.next
-    # go k steps
+    if k <= 0 or k > ll.size:
+        raise IndexError('list index out of range')
+    # go size - k steps
+    n = ll.head
     i = 1
-    n = reversed_ll.head
-    while n is not None and i < k:
+    while n is not None and i <= ll.size - k:
```
lhchavez: maybe use a `for` instead?

miguelHx: updated to use for loop

miguelHx: could you re-approve the PRs? My nit commits changed status from approved to unreviewed.

I understand that you still can't merge them due to privilege reasons, though.

PR #40, opened 2019-07-01
Miguel 2.4 - Partition [Python]

miguelHx:


Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
```diff
@@ -0,0 +1,241 @@
...
+Write code to partition a linked list around a value x,
+such that all nodes less than x come before all nodes
+greater than or equal to x. If x is contained within
+the list, the values of x only need to be after
+the elements less than x. The partition element x
+can appear anywhere in the 'right partition'; it does
+not need to appear between the left and right
+partitions.
+EXAMPLE
+Input:  3 -> 5 -> 8 -> 5 -> 10 -> 2 -> 1 [partition = 5]
```
lhchavez: in the context of Quicksort, the value x is normally called ["pivot"](https://en.wikipedia.org/wiki/Quicksort). we cannot call it "partition" since that's the name of the operation as a whole, so might as well call it "pivot". Same in the rest of the file.

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
```diff
@@ -0,0 +1,241 @@
...
+    in the linked list.
+    Order doesn't matter, as long as the values
+    < partition come before values >= partition.
+    Runtime: O(n)
+    Space Complexity: O(1)
+    :param ll: an input linked list
+    :param partition: a number to partition around
+    :return:
+    """
+    if ll.size == 1:
```
lhchavez: is this needed?

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
```diff
@@ -0,0 +1,241 @@
...
+        return True
+    # counter for keeping track of how many times
+    # we change from < partition to >= partition
+    partition_change_count = 0
+    changed = False
```
lhchavez: this looks like the [for-if anti-pattern](https://devblogs.microsoft.com/oldnewthing/20111227-00/?p=8793). you can split the loop in two to avoid it.

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
```diff
@@ -0,0 +1,241 @@
...
+    This function will take in a linked list, a
+    partition value, and will partition a linked
+    list around a value x such that all nodes
+    less than x come before all nodes greater than
+    or equal to x.  If x is in the list, then
+    the value(s) of x only need to be after the
+    elements less than x. They do not need to appear
+    in between the left and right partitions.
+    Runtime: O(n)
+    Space Complexity: O(n)
```
lhchavez: Is this optimal?

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
miguelHx: nope, removed

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
miguelHx: changed

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
miguelHx: To confirm my understanding, does it look like the for-if anti-pattern because more iterations of the loop are wasted than is necessary?

I split into two loops, but it seems that more code is duplicated.  It code duplication justifiable when it avoids an anti-pattern?

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
miguelHx: Not optimal because we can instead swap in place, leading to O(1) space complexity, correct?

lhchavez: correct. you can also move the nodes to two temporary linkedlists and then splice them together.

neither alternative would worsen the runtime complexity.

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
```diff
@@ -0,0 +1,240 @@
...
+            (
+                LinkedList(10, 9, 8, 7, 6, 5, 4, 3, 2, 1),
+                5,
+                False
+            ),
+            (
+                LinkedList(1, 2, 3, 4, 5, 6, 7, 8, 9),
+                5,
+                True
+            )
```
lhchavez:
```suggestion
            ),
            (
                LinkedList(5, 1),
                5,
                False
            ),
```

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
lhchavez: i left a comment for a missing test case. there's a bug in the implementation, and one way of fixing the bug leads to fewer duplicated code.

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
miguelHx: Wouldn't creating two temporary linked lists result in O(n) space still?  Also, since a new linked list is the output, wouldn't that mean O(n) space either way?

lhchavez: not if you move the nodes from one list to another. the list itself is just two pointers and a size.

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
miguelHx: The test case that you added LinkedList(5, 1), 5, seems to pass.

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
miguelHx: Ah, I see.

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
```diff
@@ -0,0 +1,245 @@
...
+    pivot_change_count = 0
+    n = ll.head
+    while n is not None:
+        if n.data >= pivot:
+            pivot_change_count += 1
+            break
+        n = n.next
+    while n is not None:
+        if n.data < pivot:
+            pivot_change_count += 1
```
lhchavez: `pivot_change_count` cannot be less than 1 here, so whenever this line is reached, you can simply `return False`. that means that L126 can never be more than 1, so you can change that one to `return True`. finally, you are not reading `pivot_change_count` anywhere, so you can remove any mentions of that variable.

which ultimately leads to code that is shorter, easier to understand _and_ has less repetition!

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
lhchavez: ha, i misunderstood the code. but that makes a sketch of a proof i laid out in a comment easier to argue for.

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
```diff
@@ -137,23 +151,28 @@ def partition_ll(ll: LinkedList, pivot: int) -> LinkedList:
...
-            left_partition.append(n.data)
+            left_partition.append_to_tail(n)
         else:
-            right_partition.append(n.data)
+            right_partition.append_to_tail(n)
         n = n.next
+    # last element may still be pointing
+    # to earlier elements that are less than
+    # pivot so we need to cut that link
+    right_partition.append_to_tail(None)
```
lhchavez: can you instead add a `def pop_head(self) -> Node` to avoid doing this? that way you can do

```python
left_partition.append_to_tail(ll.pop_head())
```

that way you can also avoid touching the `head` and `next` pointers.

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
```diff
@@ -137,23 +151,28 @@ def partition_ll(ll: LinkedList, pivot: int) -> LinkedList:
...
     # then, merge left and right partition lists into one linked list
-    return LinkedList(*left_partition, *right_partition)
+    left_partition.tail.next = right_partition.head
```
lhchavez: can you add a `def append(self, ll: LinkedList) -> None` to prevent the implementation details from leaking?

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
```diff
@@ -83,6 +83,18 @@ def get_node_at(self, index: int) -> Node:
...
+    def pop_head(self) -> Optional[Node]:
+        if self.head is None:
+            return self.head
+        h = self.head
+        self.head = self.head.next
+        return h
+
+    def append(self, ll: object):
+        if not isinstance(ll, LinkedList):
+            return TypeError
```
lhchavez: shouldn't this be

```suggestion
            raise TypeError(...)
```

? the `return SomeError` pattern only occurs within `__eq__` and its comparison overload friends.

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
```diff
@@ -0,0 +1,276 @@
...
+            raise IndexError('list index out of range')
+        n = self.head
+        for i in range(self.size):
+            if i == index:
+                return n
+            n = n.next
+
+    def pop_head(self) -> Optional[Node]:
+        if self.head is None:
+            return self.head
```
lhchavez: maybe raise something instead of returning `None`. it's kinda like doing `[].pop()`

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
```diff
@@ -0,0 +1,276 @@
...
+        h = self.head
+        self.head = self.head.next
+        return h
```
lhchavez: can you set `h.next = None` before returning it? that way the node is self-contained. also, you need to decrement the length of the list by one.

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
```diff
@@ -0,0 +1,276 @@
...
+    right_partition = LinkedList()  # will contain values >= pivot
+
+    n = ll.head
+    while n is not None:
+        if n.data < pivot:
+            left_partition.append_to_tail(ll.pop_head())
+        else:
+            right_partition.append_to_tail(ll.pop_head())
+        n = n.next
+    # last element may still be pointing
```
lhchavez: this should no longer be the case with the above suggestion.

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
```diff
@@ -0,0 +1,276 @@
...
+        if self.head is None:
+            return self.head
+        h = self.head
+        self.head = self.head.next
+        return h
+
+    def append(self, ll: object):
+        if not isinstance(ll, LinkedList):
+            return TypeError
+        self.tail.next = ll.head
```
lhchavez: you also need to update the size. and maybe clear out the latter list to avoid any preconditions from being violated by manipulating one of the lists without the other one knowing (since they would be sharing the Nodes and that sounds scary).

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
```diff
@@ -92,7 +92,7 @@ def pop_head(self) -> Optional[Node]:
 
     def append(self, ll: object):
         if not isinstance(ll, LinkedList):
-            return TypeError
+            raise TypeError(...)
```
lhchavez: err, i didn't mean a literal `...` ^^;; you can use `raise IndexError('pop from empty list')` just like `[].pop()`.

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
```diff
@@ -92,7 +92,7 @@ def pop_head(self) -> Optional[Node]:
 
     def append(self, ll: object):
```
lhchavez:
```suggestion
    def append(self, ll: LinkedList) -> None:
```

that way you don't even need the `isinstance`.

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
miguelHx: doesn't that defeat the purpose of having Optional[Node]?

Also, for `right_partition.append_to_tail(ll.pop_head())` this leads to an error if I don't return None.  By that line of code, the original linked list `ll` no longer has a head.

miguelHx: Disregard my comment, found out that I don't need that line of code after the while loop anymore lol

lhchavez: that totally defeats the purpose of having it return `Optional[Node]`, and that's a good thing! ultimately, why do you _need_ to return that? isn't it better to make it behave like `list.pop()`? otherwise you're going against the [principle of least astonishment](https://en.m.wikipedia.org/wiki/Principle_of_least_astonishment).

same thing for the latter: why do you _need_ that line? isn't it better to fix that problem elsewhere within this class to avoid the implementation details from leaking?

miguelHx: updated 👍

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
miguelHx: I get a `NameError: name 'LinkedList' is not defined` exception when doing this.

lhchavez: oh right, i always forget about that quirk.

```suggestion
     def append(self, ll: 'LinkedList') -> None:
```

(note the quotes).

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
lhchavez: the latter part is still missing:

```suggestion
        ll.head = None
        ll.size = 0
```

miguelHx: oops, my negligence

Python/chapter02/2.4 - Partition/miguel_2.4_soln.py
```diff
@@ -0,0 +1,275 @@
...
+        if self.head is None:
+            self.head = Node(d)
+            self.tail = self.head
+        else:
+            end = Node(d)
+            self.tail.next = end
+            self.tail = end
+        self.size += 1
+
+    def _append_node(self, n: Optional[Node] = None) -> None:
```
lhchavez: argh, should have caught this earlier: you now don't need this to be `Optional`!

miguelHx: going to update rn

miguelHx: updated. btw, thank for taking the time to review my code.  I feel like I'm becoming a better programmer than before.  I really appreciate it!

PR #41, opened 2019-07-02
Miguel 2.5 - Sum Lists [Python]

miguelHx:


Python/chapter02/2.5 - Sum Lists/miguel_2.5_soln.py
```diff
@@ -0,0 +1,285 @@
...
+        while a is not None and b is not None:
+            if a.data != b.data:
+                return False
+            # otherwise, advance both pointers
+            a = a.next
+            b = b.next
+        return a is None and b is None
+
+
+def sum_lists_forward(ll1: LinkedList, ll2: LinkedList) -> LinkedList:
```
lhchavez: why this is needed?

Python/chapter02/2.5 - Sum Lists/miguel_2.5_soln.py
```diff
@@ -0,0 +1,285 @@
...
+    # if carry > 0:
+    #     output_ll.append_to_tail(carry)
+    return output_ll
+
+
+def sum_lists(ll1: LinkedList, ll2: LinkedList) -> LinkedList:
+    """
+    Will add ll1 and ll2 where each element in the linked lists
+    represent digits.
+    Precondition: ll1 and ll2 must be the same length.
```
lhchavez: This is not in the original problem statement. Can it be removed?

Python/chapter02/2.5 - Sum Lists/miguel_2.5_soln.py
miguelHx: It's the follow up to this question, forgot to add the comment for it.  Updated now

Python/chapter02/2.5 - Sum Lists/miguel_2.5_soln.py
miguelHx: Yes.  One solution would be to have the smaller linked list be padded with 0s.

lhchavez: can you think of a way to solve it without modifying the input?

miguelHx: I'm thinking that I can traverse through the shorter linked list, add as before, then traverse through the remaining digits on the bigger list adding any leftover carries and bringing the values down.

Python/chapter02/2.5 - Sum Lists/miguel_2.5_soln.py
```diff
@@ -0,0 +1,315 @@
...
+            a = a.next
+            b = b.next
+        return a is None and b is None
+
+
+def sum_lists_forward(ll1: LinkedList, ll2: LinkedList) -> LinkedList:
+    """
+    Will add ll1 and ll2 where each element in the linked lists
+    represent digits.
+    Precondition: ll1 and ll2 must be the same length.
```
lhchavez: do you think you can do the same for this version?

Python/chapter02/2.5 - Sum Lists/miguel_2.5_soln.py
```diff
@@ -0,0 +1,315 @@
...
+                output_ll.append_to_head(carry)
+            else:
+                # add carry to previous digit
+                prev_node.data += carry
+        output_ll.append_to_tail(result)
+        n1 = n1.next
+        n2 = n2.next
+        prev_node = output_ll.tail
+
+    # if carry > 0:
```
lhchavez: i see that there are tests for this case, but how are they passing? O_o

Python/chapter02/2.5 - Sum Lists/miguel_2.5_soln.py
miguelHx: yes

Python/chapter02/2.5 - Sum Lists/miguel_2.5_soln.py
miguelHx: is there a bug in the algorithm that would or should prevent the test cases for forward sum from passing?

lhchavez: Nevermind, I was confused about the fact that the carry code was commented out, but it was buried inside the loop.

it would have probably been better to add a break or something to be able to have each chunk of code do a single logical operation: the loop should have only dealt with the core addition loop.

Python/chapter02/2.5 - Sum Lists/miguel_2.5_soln.py
```diff
@@ -0,0 +1,447 @@
...
+    # otherwise, we have non-equal lengths
+    # with or without a leading backward carry
+    # leading bw carry means the the size diff
+    # decreased.
+    # no leading bw carry means size diff stayed the same
+    if leading_bw_carry:
+        backwards_carry = 1
+    else:
+        backwards_carry = 0
+    # otherwise, as long as the diff is >= 0
```
lhchavez: This code is kinda complicated and not too easy to follow since it has a lot of edge cases :( In order to coalesce all the edge cases together, would it be simpler if you allowed yourself to mutate the output linked list before returning it? Either the values and/or the order of the nodes within the linked list.

That way you can deal with the core functionality at first (finding the first shared node, then adding both lists), and you can push the carry logic to a second pass.

miguelHx: > In order to coalesce all the edge cases together, would it be simpler if you allowed yourself to mutate the output linked list before returning it? Either the values and/or the order of the nodes within the linked list.

What do you mean by mutating the output linked list?  Do you mean after the core functionality, then I can pad the output list with 0s?  That approach would be cleaner.  The second pass would handle that.

>
It does add an extra pass, but the overall time / space complexity would be unchanged. In any case, most problems that deal with linked lists are sort of artificial: in a realistic scenario one would change the linked list to a better data structure instead of doing all these contortions, so I wouldn't feel too bad about the execution time / code complexity trade-off.

Yeah, if this was a doubly linked list, it would have been a bit more straightforward. (or rather, straightbackward ;))

lhchavez: > > In order to coalesce all the edge cases together, would it be simpler if you allowed yourself to mutate the output linked list before returning it? Either the values and/or the order of the nodes within the linked list.
> 
> What do you mean by mutating the output linked list? Do you mean after the core functionality, then I can pad the output list with 0s? That approach would be cleaner. The second pass would handle that.

Not sure how padding the output list with zeroes would help, but if it leads to simpler code, go for it. There's also the option of adding stuff to the nodes without taking care of the carry and deferring the handling of carry to afterwards.

Whatever modifications you do to the output list are implementation details and are fair game, as long as the return value is correct.

> 
> > It does add an extra pass, but the overall time / space complexity would be unchanged. In any case, most problems that deal with linked lists are sort of artificial: in a realistic scenario one would change the linked list to a better data structure instead of doing all these contortions, so I wouldn't feel too bad about the execution time / code complexity trade-off.
> 
> Yeah, if this was a doubly linked list, it would have been a bit more straightforward. (or rather, straightbackward ;))

you can invert a singly-linked list in O(n) time O(1) space if you need to.

miguelHx: Would it be fair game to make reversed copies of the input linked lists and then call the sum_lists function on them?

lhchavez: totally!

miguelHx: Wow lol I should have done that from the beginning but I thought it was too simple

lhchavez: simplicity is ranked pretty high in the list of desirable properties of software. but simplicity is not always free, since it sometimes comes with a trade-off of time or space. you often need to at least enumerate all your options and choose whatever is best in that instance.

Python/chapter02/2.5 - Sum Lists/miguel_2.5_soln.py
```diff
@@ -149,96 +149,25 @@ def sum_lists_forward(ll1: LinkedList, ll2: LinkedList) -> LinkedList:
...
-    # n_shorter will point to the beginning of shorter_ll
-    # (9) -> 9
-    for i in range(length_diff):
-        n_longer = n_longer.next
+    n1 = ll1.head
+    n2 = ll2.head
+    # reverse both lists
+    reversed_ll1 = LinkedList()
+    reversed_ll2 = LinkedList()
+    while n1 is not None:
```
lhchavez: can you create a function for this?

Python/chapter02/2.5 - Sum Lists/miguel_2.5_soln.py
```diff
@@ -149,96 +149,25 @@ def sum_lists_forward(ll1: LinkedList, ll2: LinkedList) -> LinkedList:
...
-            output_ll.head.data = result
-        else:
-            # if backwards carry is 0, bring corresponding digit down
-            result = n_longer.data
-            output_ll.append_to_head(result)
-        length_diff -= 1
-    # last backwards carry
-    if backwards_carry == 1:
-        output_ll.append_to_head(backwards_carry)
+    n = reversed_result.head
```
lhchavez: i'm a bit on the fence about this extra copy. there's a way to reverse a linked list in O(1) space, but it's a bit more complex. and linked lists are sort of boring, so up to you to create an additional version of the linked list reversing function that destructively reverses the output list (you can add it to the LinkedList class and can do it "in-place").

Python/chapter02/2.5 - Sum Lists/miguel_2.5_soln.py
miguelHx: yes

Python/chapter02/2.5 - Sum Lists/miguel_2.5_soln.py
miguelHx: updated

PR #42, opened 2019-07-16
Miguel 2.6 - Palindrome [Python]

miguelHx:


Python/chapter02/2.6 - Palindrome/miguel_2.6_soln.py
```diff
@@ -0,0 +1,218 @@
...
+    """
+    Given a linked list, this function will check if the
+    linked list is a palindrome.
+    A palindrome is a word or phrase that is the same
+    forwards and backwards.
+    Since I will use integer nodes for the linked list,
+    we will be checking if the sequence of numbers
+    in a linked list is a palindrome.
+    Runtime:  O(n)
+    Space Complexity:  O(n)
```
lhchavez: is this optimal?

miguelHx: I think for runtime yes, but space complexity can be constant if I compare each element in the linked list with its complement (length - I - 1) up until the halfway point without creating a new reversed linked list.

lhchavez: yes, runtime you cannot do better than O(n) since you need to at least inspect every single element. but space can be improved as you say.

miguelHx: ok, going to write a new function for that.

Python/chapter02/2.6 - Palindrome/miguel_2.6_soln.py
```diff
@@ -154,6 +155,29 @@ def reverse_linked_list(ll: LinkedList) -> LinkedList:
     return output_ll
 
 
+def palindrome_constant_space(ll: LinkedList) -> bool:
+    """
+    Same as the palindrome function below, but
+    with constant space
+    Runtime:  O(n^2)
```
lhchavez: is this optimal?

Python/chapter02/2.6 - Palindrome/miguel_2.6_soln.py
```diff
@@ -154,6 +155,29 @@ def reverse_linked_list(ll: LinkedList) -> LinkedList:
...
+        for j in range(i, ll.size - i - 1):
+            n2 = n2.next
+        # if not the same, then not a palindrome
+        if n2.data != curr_data:
+            return False
+        n1 = n1.next
+    return True
+
+
 def palindrome(ll: LinkedList) -> bool:
```
lhchavez: forgot this one in the previous iteration: nit: call this `is_palindrome` since it returns a boolean. that way you can more easily guess what itś going to do.

Python/chapter02/2.6 - Palindrome/miguel_2.6_soln.py
miguelHx: yes

lhchavez: can you justify your answer?

miguelHx: The outer loop will be doing n/2 steps.  As we get closer to n/2, there will be less steps (steps = advancing n2 pointer).  The number of steps at each iteration is n - 2*i - 1.  The sequence will look like this:  (n-1) + (n-3) + (n-5) + ... + 5 + 3 + 1.  So, the number of steps will be the sum of that sequence.  We can add the same sequence to itself and then rearrange to get (n + n + n + ... + n + n + n), and since we have n/2 terms, we get n * n/2 = n^2/2.  But we need to divide by 2 again to undo the adding sequence to itself part, which results in n^2 / 4.  I could be wrong, but this is my reasoning.

Python/chapter02/2.6 - Palindrome/miguel_2.6_soln.py
miguelHx: updated

Python/chapter02/2.6 - Palindrome/miguel_2.6_soln.py
miguelHx: Oh wait...

miguelHx: I answered the wrong question, you asked if this is optimal, not to justify my runtime analysis.

lhchavez: that is a decent justification of why the current code is O(n^2), but not whether it's optimal!

i'll ask in a different way: can you think of an O(1) space/time way to compare the head and tail of the linked list for equality?

miguelHx: O(1) space and O(1) time way?  Or do you mean O(1) but better than O(n^2) time way?

miguelHx: Head and tail can be compared in constant time, but I'm not sure about the rest.  If this was a doubly linked list, I can get a O(n) runtime with O(1) space.

lhchavez: the former.

miguelHx: Well for the first and last elements we can access head and tail in constant time.  However, for the middle elements, I don't think the complement can be accessed in constant time unless we had a doubly linked list and 'popped' the head and tail each time.

lhchavez: you already have a method that can reverse the linked list in-place. what if you only reversed half?

miguelHx: right, I forget that creating a new LinkedList() doesn't always use n space.  Solution updated

Python/chapter02/2.6 - Palindrome/miguel_2.6_soln.py
```diff
@@ -0,0 +1,256 @@
...
+    Runtime:  O(n)
+    Space Complexity:  O(1)
+    :param ll: an input linked list
+    :return: true if ll is a palindrome, false otherwise
+    """
+    # reverse half of the linked list
+    temp_ll1 = LinkedList()
+    temp_ll2 = LinkedList()
+    n = ll.head
+    for _ in range(math.ceil(ll.size/2)):
```
lhchavez:
```suggestion
    for _ in range((ll.size + 1) // 2):
```

this is the more pythonic way of doing this.

Python/chapter02/2.6 - Palindrome/miguel_2.6_soln.py
```diff
@@ -0,0 +1,256 @@
...
+    """
+    output_ll = LinkedList()
+    n = ll.head
+    while n is not None:
+        output_ll.append_to_head(n.data)
+        n = n.next
+    return output_ll
+
+
+def is_palindrome_constant_space(ll: LinkedList) -> bool:
```
lhchavez: can you change the test for this function to validate that the contents of the linked list are "preserved"? (as in, make a copy beforehand, and then compare the copy with the input list to make sure the values of the entries are the same).

miguelHx: updated

Python/chapter02/2.6 - Palindrome/miguel_2.6_soln.py
miguelHx: updated

PR #43, opened 2019-07-22
Miguel 2.7 - Intersection [Python]

miguelHx:


Python/chapter02/2.7 - Intersection/miguel_soln_2.7.py
```diff
@@ -0,0 +1,234 @@
...
+            n2 = n2.next
+        n1 = n1.next
+    return None
+
+
+class TestIntersection(unittest.TestCase):
+
+    def setUp(self):
+        shared_structures = [
+            (
```
lhchavez: consider using [`typing.NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple).

Python/chapter02/2.7 - Intersection/miguel_soln_2.7.py
```diff
@@ -0,0 +1,234 @@
...
+                LinkedList(1, 2, 3)
+            ),
+            (
+                LinkedList(66, 2, 12, 35),
+                LinkedList(79, 19, 23, 24)
+            ),
+            (
+                LinkedList(),
+                LinkedList(1, 2, 3, 4)
+            )
```
lhchavez: maybe also add two lists with the same values

Python/chapter02/2.7 - Intersection/miguel_soln_2.7.py
```diff
@@ -0,0 +1,234 @@
...
+            a = a.next
+            b = b.next
+        return a is None and b is None
+
+
+def intersection(ll1: LinkedList, ll2: LinkedList) -> Optional[Node]:
+    """
+    This function will determine if ll1 and ll2 intersect.
+    The intersection is defined based on reference, not value.
+    Runtime:  O(n^2)
```
lhchavez: is this optimal?

Python/chapter02/2.7 - Intersection/miguel_soln_2.7.py
miguelHx: updated

Python/chapter02/2.7 - Intersection/miguel_soln_2.7.py
miguelHx: updated

Python/chapter02/2.7 - Intersection/miguel_soln_2.7.py
```diff
@@ -0,0 +1,245 @@
...
+                LinkedList(1, 2),
+                LinkedList(4, 5),
+                Node(3),
+                LinkedList(9, 8, 7, 6, 7, 6, 7, 9),
+            )
+        ]
+
+        self.intersection_test_cases = []
+
+        for ll1_seg1, ll1_seg2, shared_node, ll2 in shared_structures:
```
lhchavez: now you don't need these abbreviated variables!

Python/chapter02/2.7 - Intersection/miguel_soln_2.7.py
miguelHx: no, we can loop through first linked list, store references in a set.  Then loop through second list, checking if each node reference is in the set.
Would be O(n) Time with O(n) space

Python/chapter02/2.7 - Intersection/miguel_soln_2.7.py
miguelHx: updated

PR #44, opened 2019-08-01
Jesus 1.1 - is_unique [Python] 

jsanchez254:


Python/chapter01/1.1 - Is Unique/jesus_1.1.sol.py
```diff
@@ -0,0 +1,22 @@
+import unittest
+
+
+def is_unique(s: str) -> bool:
```
lhchavez: Can you add a [docstring](https://www.python.org/dev/peps/pep-0257/)? In the docstring, please document what the runtime+space complexity is.

lhchavez: good job on adding typing information! :D

Python/chapter01/1.1 - Is Unique/jesus_1.1.sol.py
```diff
@@ -0,0 +1,22 @@
...
+    for x in s:
+        if s.count(x) > 1:
+            return False
+    return True
+
+class TestIsUnique(unittest.TestCase):
+
+    def test_is_unique(self):
+        self.assertEqual(is_unique('aaaaa'), False)
```
lhchavez: there are better assertions you can use: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertTrue

```suggestion
        self.assertFalse(is_unique('aaaaa'))
```

same below

Python/chapter01/1.1 - Is Unique/jesus_1.1.sol.py
```diff
@@ -0,0 +1,22 @@
+import unittest
```
lhchavez: Would it be possible for you to add a module docstring with the problem statement? Not all (volunteer!) reviewers have a copy of the book.

Python/chapter01/1.1 - Is Unique/jesus_1.1.sol.py
```diff
@@ -0,0 +1,22 @@
...
+
+def is_unique(s: str) -> bool:
+    for x in s:
+        if s.count(x) > 1:
+            return False
+    return True
+
+class TestIsUnique(unittest.TestCase):
+
+    def test_is_unique(self):
```
lhchavez: this might be okay for now, but for more complex tests, Golang has this wonderful pattern called [table driven tests](https://github.com/golang/go/wiki/TableDrivenTests), where you can restructure the code the following way:

```python
for s, expected in [
  ('aaaaa', False),
  ('abc', True),
  ...
  ('a', True),
]:
    self.assertEqual(is_unique(s), expected, msg=s)
```

to cut down on repetition.

for even further de-duplication, you can split the test in positive and negative tests:

```python
def test_is_unique(self):
  for s in [
    'abc',
    ....
    '',
  ]:
    self.assertTrue(is_unique(s), msg=s)

def test_is_not_unique(self):
  for s in [
    'aaaaa',
    ....
    'abcda',
  ]:
    self.assertFalse(is_unique(s), msg=s)
```

PR #45, opened 2019-08-02
Miguel 2.8 - Loop Detection [Python]

miguelHx:


Python/chapter02/2.8 - Loop Detection/miguel_soln_2.8.py
```diff
@@ -0,0 +1,240 @@
...
+    A linked list is 'circular' when a node's
+    next pointer points to an earlier node, so
+    as to make a loop in the linked list.
+    Runtime:  O(n^2)
+    Space Complexity:  O(1)
+    :param ll: an input linked list
+    :return: the corrupt node or None
+    """
+    n1 = ll.head
+    for i in range(ll.size+1):
```
lhchavez: plot twist: with circular linked lists, the size is not defined (the problem statement does say it's corrupted). can you think of a way to solve this without relying on this value?

miguelHx: updated

Python/chapter02/2.8 - Loop Detection/miguel_soln_2.8.py
```diff
@@ -171,20 +171,18 @@ def loop_detection(ll: LinkedList) -> Optional[Node]:
     A linked list is 'circular' when a node's
     next pointer points to an earlier node, so
     as to make a loop in the linked list.
-    Runtime:  O(n^2)
-    Space Complexity:  O(1)
+    Runtime:  O(n)
+    Space Complexity:  O(n)
```
lhchavez: nice! this is much better _and_ simpler to understand.

stretch goal: there exists an O(1) space solution that does not rely on size.

miguelHx: updated, O(1) space with O(n^2) runtime

Python/chapter02/2.8 - Loop Detection/miguel_soln_2.8.py
```diff
@@ -164,6 +164,31 @@ def __eq__(self, other: object):
...
+    cycle in the input linked list.
+    A linked list is 'circular' when a node's
+    next pointer points to an earlier node, so
+    as to make a loop in the linked list.
+    Runtime:  O(n^2)
+    Space Complexity:  O(1)
+    :param ll: an input linked list
+    :return: the corrupt node or None
+    """
+    if ll.head and ll.head.next is ll.head:
```
lhchavez: there doesn't seem to be a test case for this.

Python/chapter02/2.8 - Loop Detection/miguel_soln_2.8.py
```diff
@@ -164,6 +164,31 @@ def __eq__(self, other: object):
...
+        return ll.head
+    n1 = ll.head
+    while n1 is not None:
+        n2 = ll.head
+        while n2 is not n1:
```
lhchavez: what happens if you add a test case with a very long chain of elements and a very long cycle? (say, 10 items each)

also, can you modify the test generation such that you only manipulate the `.next` pointers instead of using the append functions? there's a bug here that's not being caught by the tests :S

Python/chapter02/2.8 - Loop Detection/miguel_soln_2.8.py
miguelHx: that is for a single node linked list that has a next pointer pointing to itself

miguelHx: The test method for that is `def test_loop_detection_single_node_ll`

lhchavez: ah somehow i missed that. sorry about the noise.

miguelHx: no worries

Python/chapter02/2.8 - Loop Detection/miguel_soln_2.8.py
miguelHx: long cycle test case added.

I do not see the bug that is not being caught by the tests.  Could you give me another hint?

lhchavez: after more careful inspection, it was just me misreading the code u__u.

i was misinterpreting what `n1` and `n2` meant. can you add a few comments about how this all works and maybe also rename those variables to avoid future readers from also being confused?

Python/chapter02/2.8 - Loop Detection/miguel_soln_2.8.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+def loop_detection_const_space(ll: LinkedList) -> Optional[Node]:
+    """
+    This function will determine if there is a
+    cycle in the input linked list.
+    A linked list is 'circular' when a node's
+    next pointer points to an earlier node, so
+    as to make a loop in the linked list.
+    Runtime:  O(n^2)
```
lhchavez: extra bonus points: there exists an O(n) time, O(1) space algorithm!

Python/chapter02/2.8 - Loop Detection/miguel_soln_2.8.py
miguelHx: Yes, good idea!

miguelHx: updated

Python/chapter02/2.8 - Loop Detection/miguel_soln_2.8.py
miguelHx: I think I got it

miguelHx: Although the solution I just updated only works if the linked list's tail's next pointer points to a previous node.

Python/chapter02/2.8 - Loop Detection/miguel_soln_2.8.py
```diff
@@ -0,0 +1,310 @@
...
+    next pointer points to an earlier node, so
+    as to make a loop in the linked list.
+    Runtime:  O(n)
+    Space Complexity:  O(1)
+    :param ll: an input linked list
+    :return: the corrupt node or None
+    """
+    curr_node = ll.head
+    while curr_node is not None:
+        if curr_node is ll.tail.next:
```
lhchavez: note that you cannot rely on anything except for each node's `.next` pointer. this just so happens to pass the tests because that's the way you constructed the tests' inputs.

miguelHx: updated

Python/chapter02/2.8 - Loop Detection/miguel_soln_2.8.py
```diff
@@ -171,16 +171,20 @@ def loop_detection_linear_time_const_space(ll: LinkedList):
...
-    while curr_node is not None:
-        if curr_node is ll.tail.next:
-            return curr_node
-        curr_node = curr_node.next
+    slow_ptr = ll.head
+    fast_ptr = ll.head
+    while slow_ptr and fast_ptr and fast_ptr.next:
+        slow_ptr = slow_ptr.next
+        fast_ptr = fast_ptr.next.next
+        if slow_ptr is fast_ptr or slow_ptr is fast_ptr.next:
```
lhchavez: you only need one of these comparisons. Adding both makes it harder to reason about.

miguelHx: For the test case 1 -> 2 -> 3 -> 4 -> 5 -> 3
leaving out the second comparison leads to the slow and fast pointer meeting up at the node containing 4, which is not the start of the loop and caused the first test case to fail.

lhchavez: i'm pretty sure you can construct a test case that makes it fail _even_ with the duplicate test. say, if the loop is long enough, it'll detect the loop somewhere in the middle.

what this means is that the cycle-finding algorithm only lets you know whether there is a cycle and _a_ node within it. you need additional logic if you want the very first.

lhchavez: (and having said that, can you find such a test case?)

miguelHx: Oh ok you are right,
Yes I constructed such a test case.
Going to add the logic for finding the first node in the cycle

lhchavez: in general, every time you add complexity (mostly in the form of one-off tests and branches) to your algorithm, you need to analyze whether those one-off tests are needed (maybe they are covered by a more general case), and whether they are fixing a real issue or just papering over one narrow set of cases.

This is what trips 80% of the interviewees in my experience.

miguelHx: updated

Python/chapter02/2.8 - Loop Detection/miguel_soln_2.8.py
```diff
@@ -183,8 +183,20 @@ def loop_detection_linear_time_const_space(ll: LinkedList):
     while slow_ptr and fast_ptr and fast_ptr.next:
         slow_ptr = slow_ptr.next
         fast_ptr = fast_ptr.next.next
-        if slow_ptr is fast_ptr or slow_ptr is fast_ptr.next:
-            return slow_ptr
+        if slow_ptr is not fast_ptr:
```
lhchavez: would it make sense to `break` upon reaching the opposite condition? or something like

```python
if fast_ptr is None:
    return None
if slow_ptr is fast_ptr:
    # if we get here, then there is a cycle.
    break
# ... the rest, to avoid nesting
```

Python/chapter02/2.8 - Loop Detection/miguel_soln_2.8.py
```diff
@@ -183,8 +183,20 @@ def loop_detection_linear_time_const_space(ll: LinkedList):
...
+            continue
+        # if we get here, then there is a cycle.
+        # advance one of fast or slow pointers
+        # and a pointer that starts in the
+        # beginning, by one until they match.
+        # they will end at the beginning of
+        # the cycle.
+        p = ll.head
+        while p and slow_ptr:
```
lhchavez: given that there is a cycle, `p` and `slow_ptr` will never be `None`, right? you can instead do

```python
while p is not slow_ptr:
    p = p.next
    slow_ptr = slow_ptr.next
return p
```

Python/chapter02/2.8 - Loop Detection/miguel_soln_2.8.py
```diff
@@ -183,8 +183,20 @@ def loop_detection_linear_time_const_space(ll: LinkedList):
     while slow_ptr and fast_ptr and fast_ptr.next:
         slow_ptr = slow_ptr.next
         fast_ptr = fast_ptr.next.next
-        if slow_ptr is fast_ptr or slow_ptr is fast_ptr.next:
-            return slow_ptr
+        if slow_ptr is not fast_ptr:
+            continue
+        # if we get here, then there is a cycle.
+        # advance one of fast or slow pointers
```
lhchavez: This comment is mostly explaining what the code does, which is not super useful. It also doesn't _quite_ reflect the code, since you say "advance one of fast or slow pointers" but you chose to advance the slow pointer.

Would it be possible to explain _why_ / _how_ just advancing the pointers at the same speed will always make them both align at the beginning of the cycle?

As a hint: try to decompose the position of the fast and slow pointers in terms of the length of the list leading to the cycle, the length of the cycle, and the offset from the start of the cycle. Sprinkle in some modular arithmetic and you should be able to come up with a rough proof of correctness.

PR #46, opened 2019-10-11
Added a program to remove duplicates from an unsorted singlely liked …

aditya9061:
…list.

### DIFF
--- a/JavaScript/chapter02/2.1 - Remove Dups/solution.js
+++ b/JavaScript/chapter02/2.1 - Remove Dups/solution.js
@@ -1,0 +1,93 @@
+class node{
+	
+	constructor(data){
+		
+		this.data=data;
+		this.next=null;
+
+	}
+}
+
+
+class LinkedList{
+	
+	constructor(){
+		this.head=null;
+	}
+
+//Function to insert a node in the singlely linkedlist.
+	insert=(data)=>{
+		
+		var newnode=new node(data);
+		if(head===null){
+			this.head=newnode;
+		}
+
+		else{
+			
+			var temp=this.head;
+			while(temp.next!==null){
+				temp=temp.next;			
+			}		
+				
+			temp=newnode;
+		}	
+	}
+
+//Function to print the singlely linkedlist.
+	printList=()=>{
+		var temp=this.head;
+		var list;
+		while(temp!==null){
+			list+=temp.data+"->";
+			temp=temp.next;		
+		}
+		list+="null";
+		console.log(list);
+	}
+
+	
+
+//Function to remove dupliicate nodes from the singlely linkedlist.
+	removeDuplicates=()=>{
+	
+		var temp1,temp2,duplicate;
+	
+		temp1=this.head;
+		temp2=temp1;
+		while(temp1!==null || temp.next!==null){
+			
+			while(temp2.next!==null){
+				if(temp1.data===temp2.next.data){
+					duplicate=temp2.next;
+					temp2.next=temp2.next.next;
+					delete(duplicate);
+				}
+				else{
+					temp2=temp2.next;				
+				}			
+			}
+			temp1=temp1.next;
+		}
+	
+	}
+
+}
+
+//main function.
+/*
+Linkedlist list;
+list.insert(0);
+list.insert(1);
+list.insert(2);
+list.insert(3);
+list.insert(1);
+list.insert(4);
+list.insert(4);
+console.log("Original List with DUplicates:");
+list.printList();
+list.removeDuplicates();
+console.log("List free of duplicates:");
+list.printList();
+*/
+


JavaScript/chapter02/2.1 - Remove Dups/solution.js
```diff
@@ -0,0 +1,87 @@
+class node{
```
lhchavez: be consistent with how you name things: you use CamelCase for `class LinkedList` below.

JavaScript/chapter02/2.1 - Remove Dups/solution.js
```diff
@@ -0,0 +1,87 @@
...
+		list+="null";
+		console.log(list);
+	}
+
+	
+
+//Function to remove dupliicate nodes from the singlely linkedlist.
+	removeDuplicates(){
+	
+		var temp1,temp2;
```
lhchavez: nowadays the use of `var` is discouraged due to the lack of block scoping and them being hoisted.

it is recommended that declarations use `const` if possible. otherwise, you should use `let` and declare things in the same line as they are assigned a value:

```javascript
let currentNode = this.head;
```

also, try to give more descriptive names to your variables so that when you go back and read the code in a few weeks/months, you know what they are.

JavaScript/chapter02/2.1 - Remove Dups/solution.js
```diff
@@ -0,0 +1,87 @@
...
+					temp2=temp2.next;				
+				}			
+			}
+			temp1=temp1.next;
+		}
+	
+	}
+
+}
+//main function.
```
lhchavez: can you uncomment the code so that you know when the tests are passing?

JavaScript/chapter02/2.1 - Remove Dups/solution.js
aditya9061: Thanks for the inputs, will add them in the code.

JavaScript/chapter02/2.1 - Remove Dups/solution.js
aditya9061: Sure!

JavaScript/chapter02/2.1 - Remove Dups/solution.js
aditya9061: Okay,thanks.

JavaScript/chapter02/2.1 - Remove Dups/solution.js
aditya9061: As suggested, I tried to implement all the changes (including running prettier).

JavaScript/chapter02/2.1 - Remove Dups/solution.js
```diff
@@ -0,0 +1,71 @@
...
+    while (currentNode) {
+      list += currentNode.data + "->";
+      currentNode = currentNode.next;
+    }
+    list += "null";
+    console.log(list);
+  }
+
+  //Function to remove dupliicate nodes from the singlely linkedlist.
+  removeDuplicates() {
```
lhchavez: what is the runtime/space complexity of this function?

JavaScript/chapter02/2.1 - Remove Dups/solution.js
```diff
@@ -0,0 +1,71 @@
...
+        } else {
+          currentNode2 = currentNode2.next;
+        }
+      }
+      currentNode1 = currentNode1.next;
+    }
+  }
+}
+//main function.
+var list1 = new LinkedList();
```
lhchavez: I'm trying to make it easier to know if your code is correct, so I added support for [mocha](https://mochajs.org/) in https://github.com/techqueria/data-structures-and-algorithms/pull/47

Can you change the code so that it follows the pattern found in https://github.com/techqueria/data-structures-and-algorithms/blob/aa70628efaf99a825115cea3c1008a4ef4276816/JavaScript/chapter01/1.1%20-%20Is%20Unique/solution.js ? you might need to change `printList()` in favor of [`toString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString) so that you can use `assert.equal(list.toString(), '...');`

JavaScript/chapter02/2.1 - Remove Dups/solution.js
```diff
@@ -0,0 +1,71 @@
+class Node {
```
lhchavez: i forgot to notice this before, but can you add your solution to a file called `aditya9061.js`?

JavaScript/chapter02/2.1 - Remove Dups/solution.js
aditya9061: The time complexity for removeDuplicates function is O(n²) and space complexity is O(1).
And time complexity for printList is O(n) and space complexity is O(1).

JavaScript/chapter02/2.1 - Remove Dups/solution.js
aditya9061: Sure I will make the changes soon

JavaScript/chapter02/2.1 - Remove Dups/solution.js
lhchavez: is there a way to make it faster?

JavaScript/chapter02/2.1 - Remove Dups/solution.js
aditya9061: Maybe by using recursion time complexity can be reduce.

JavaScript/chapter02/2.1 - Remove Dups/solution.js
aditya9061: Maybe by using recursion it can be made faster

lhchavez: what would the space/time complexity be with recursion?

aditya9061: O(n)

lhchavez: can you expand on that? usually with recursive algorithms, analysis becomes complicated because of the fact that computation across method calls need to be accounted for.

JavaScript/chapter02/2.1 - Remove Dups/solution.js
lhchavez: that change has been merged now!

PR #51, opened 2019-11-09
Support Travis CI

lhchavez:
This change adds a .travis.yml file that will automatically validate
that all unit tests pass.


lhchavez:
This will look like this: https://travis-ci.org/lhchavez/data-and-algorithms/builds/609675309

PR #53, opened 2020-07-30
avc278 - 1.01 - Javascript

avc278:
Added my two solutions to the Chapter 1 Problem 1 folder.
First solution solves the problem with the use of additional data structures.
Second solution solves the problem without any additional data structures.

### DIFF
--- a/JavaScript/chapter01/p01_is_unique/avc278.js
+++ b/JavaScript/chapter01/p01_is_unique/avc278.js
@@ -1,0 +1,59 @@
+// Is Unique: Implement an algorithm to determine if a string has all unique characters.
+// What if you cannot used additional data structures?
+
+const assert = require("assert");
+
+/**
+ * Checks if a string is composed of all unique characters
+ * @param  {string} str input string to check
+ * @return {bool}       Whether the input string contains unique characters
+ *
+ * We iterate through the input str of length N. Operations within this loop are O(1), so no additional time added.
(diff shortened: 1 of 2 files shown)


JavaScript/chapter01/p01_is_unique/avc278.js
```diff
@@ -0,0 +1,59 @@
...
+// What if you cannot used additional data structures?
+
+const assert = require("assert");
+
+/**
+ * Checks if a string is composed of all unique characters
+ * @param  {string} str input string to check
+ * @return {bool}       Whether the input string contains unique characters
+ *
+ * We iterate through the input str of length N. Operations within this loop are O(1), so no additional time added.
```
lhchavez, line 11: FYI: contrary to popular belief, hashtable / hashmap operations for non-integers are _not_ O(1)!

The reason is that before any operation in the data structure, the keys need to be hashed. for integers, this is effectively an O(1) operation (at most an integer division), but for other more complex types (like strings), this is O(l), where `l` is the length of the string.

Most hash algorithms are wicked fast, but they should not be considered O(1) for correct complexity analysis.

... but having said all of that, in this case each individual character is at most one 16-bit unicode code point per iteration, so in this particular case it is O(1).

avc278: Very interesting. I'll have to keep that in mind for the future. Thanks!

PR #54, opened 2020-07-31
avc278 - 1.02 - Javascript

avc278:
Added my solution to the Chapter 1 Problem 2 folder.

### DIFF
--- a/JavaScript/chapter01/p02_check_perm/avc278.js
+++ b/JavaScript/chapter01/p02_check_perm/avc278.js
@@ -15,8 +15,7 @@ const assert = require("assert");
  * At this point, we can get the keys from the frequency objects we created, which is an O(N) operation for space.
  * If the lengths of `aKeys` and `bKeys` are not the same, we know they cannot be permutations and we exit early.
  * Otherwise, we can continue to iterate through the keys. We need to make sure that the frequencies of aKeys exist in
- * both aFreq and bFreq AND that the frequencies of bKeys exist in aFreq and bFreq. It is not enough to check that one
- * exists in both and not check the other set of keys. If at any point, they do not match, we can return false.
+ * both aFreq and bFreq. If at any point, they do not match, we can return false.
  *
  * If we haven't returned false up to this point, it's safe to return true.
  *
@@ -24,16 +23,15 @@ const assert = require("assert");
  * Space:   O(N)
  *
  */
-const checkPermutation = (A, B) => {
+const checkPermutation1 = (A, B) => {
   if (A.length !== B.length) return false;
   const aFreq = {};
   const bFreq = {};
-  for (let i = 0; i < A.length; i++) {
-    aFreq[A[i]] = aFreq[A[i]] || 0;
-    aFreq[A[i]] += 1;
-
-    bFreq[B[i]] = bFreq[B[i]] || 0;
-    bFreq[B[i]] += 1;
+  for (const a of A) {
+    aFreq[a] = (aFreq[a] || 0) + 1;
+  }
+  for (const b of B) {
+    bFreq[b] = (bFreq[b] || 0) + 1;
   }
 
   const aKeys = Object.keys(aFreq);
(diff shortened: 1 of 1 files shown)


JavaScript/chapter01/p02_check_perm/avc278.js
```diff
@@ -0,0 +1,71 @@
...
+ * @return {bool}     Whether the input strings are permutations of one another
+ *
+ * Throughout the problem, we set up a couple ways to early exit. At the start, if both strings are not of the same
+ * length, we already know they cannot be permutations of each other, so we return false.
+ * Otherwise, we can continue and set up a mapping between the elements in A and B, and their frequencies.
+ * This is an O(N) operation with respect to time and space where N is the length of the string A.
+ * At this point, we can get the keys from the frequency objects we created, which is an O(N) operation for space.
+ * If the lengths of `aKeys` and `bKeys` are not the same, we know they cannot be permutations and we exit early.
+ * Otherwise, we can continue to iterate through the keys. We need to make sure that the frequencies of aKeys exist in
+ * both aFreq and bFreq AND that the frequencies of bKeys exist in aFreq and bFreq. It is not enough to check that one
```
lhchavez, line 18: > It is not enough to check that one exists in both and not check the other set of keys.

Can this statement be elaborated? Is there an example where it is not enough?

JavaScript/chapter01/p02_check_perm/avc278.js
```diff
@@ -0,0 +1,71 @@
...
+ * Space:   O(N)
+ *
+ */
+const checkPermutation = (A, B) => {
+  if (A.length !== B.length) return false;
+  const aFreq = {};
+  const bFreq = {};
+  for (let i = 0; i < A.length; i++) {
+    aFreq[A[i]] = aFreq[A[i]] || 0;
+    aFreq[A[i]] += 1;
```
lhchavez, line 33: it _might_ be a bit more idiomatic to do

```javascript
for (const a of A) {
  aFreq[a] = (aFreq[a] || 0) + 1;
}
for (const b of B) {
  bFreq[b] = (bFreq[b] || 0) + 1;
}
```

avc278: Hmm yeah that definitely seems a lot more clean to me. That change will be reflected in the next commit

JavaScript/chapter01/p02_check_perm/avc278.js
avc278: Hmm yeah, so initially when thinking about this problem, I was thinking about some math class I took on set theory where if you want to prove that set A and set B are identical, you need to show that set A is a subset of B, and B is a subset of A.
So the initial thought process was
```
aFreq = {
  'a': 1,
  'b': 1,
  'c': 2,
}
bFreq = {
  'a': 1,
  'b': 1,
  'c': 2,
  'd': 1,
}
```
If we only check the frequencies of A match with the frequencies of B, then we should be good to go, HOWEVER, now that I'm thinking about it some more, we actually don't need to check that as this type of scenario will never happen since we already have checks in place that will return false early on.

I'm going to remove that part in the next commit

JavaScript/chapter01/p02_check_perm/avc278.js
```diff
@@ -0,0 +1,98 @@
...
+  if (A.length !== B.length) return false;
+
+  A = A.split("").sort().join("");
+  B = B.split("").sort().join("");
+
+  return A === B;
+};
+
+describe(module.filename, () => {
+  it("should return false on input strings not of the same size", () => {
```
lhchavez, line 74: minor optional note: it could be better to use [parameterized tests](https://mochajs.org/#dynamically-generating-tests) to avoid repetition:

```javascript
  const checkPermutations = [checkPermutation1, checkPermutation2];
  checkPermutations.forEach((checkPermutation) => {
    describe(checkPermutation.name, () => {
      it("should return false on input strings not of the same size", () => {
        assert.ok(!checkPermutation("hello", "hi"));
      });
      // ...
    });
  });
```

avc278: Oh interesting. I'll take a look at it in a bit and update my tests

PR #55, opened 2020-07-31
Change the CI from Travis to GitHub Actions

lhchavez:
Now that we run a bunch of GitHub actions, it's probably better to be
consistent and also run the unit tests as a GitHub Action.


lhchavez:
argh, forgot that GitHub requires the first PR that introduces Actions to be in the same repo. brb.

PR #58, opened 2020-08-01
avc278 - 1.03 - Javascript

avc278:
Added my solution to the Chapter 1 Problem 3 folder.

### DIFF
--- a/JavaScript/chapter01/p03_urlify/avc278.js
+++ b/JavaScript/chapter01/p03_urlify/avc278.js
@@ -1,0 +1,50 @@
+// URLify: Write a method to replace all spaces in a string with '%20'. You may assume that the string has sufficient
+// space at the end to hold the additional characters, and that you are given the "true" length of the string.
+
+const assert = require("assert");
+
+/**
+ * URLifies an input string by replacing ' ' with '%20' and returning the modified string
+ * @param  {string} str        input string to modify
+ * @param  {int}    trueLength the "true" length of the str without trailing spaces
+ * @return {string}            output URLified string
+ *
+ * Since we cannot mutate a string in javascript, we need to use an external data structure, like an array.
+ * Creating a new data structure takes up O(N) space, where N is the length of the str.
+ * We can loop through the array from the end of the true length to the start, while keeping count of two indexes.
+ * If we encounter a space within the parameters of the word's true length, we can overwrite the element at that index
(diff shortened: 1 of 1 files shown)


JavaScript/chapter01/p03_urlify/avc278.js
```diff
@@ -0,0 +1,50 @@
...
+ * Alteranatively, we just overwrite the index of the str length counter with the index of the true length counter,
+ * and decrement the index by 1, which is also just an O(1) operation.
+ * When we're done traversing the array, we can simply return the joined array to return a string.
+ *
+ * Runtime: O(N)
+ * Space:   O(N)
+ *
+ */
+const urlify = (str, trueLength) => {
+  str = str.split("");
```
lhchavez, line 26: TL;DR: consider creating a new variable name for this.

small maintainability comment: although JavaScript does not care a bit about types, this is allowed by the language, BUT it's hard for humans to track what the contents of the variable should be (an excerpt from the journey in the mind of a random developer looking at the code: "the doc comment said it was a string, and it's being treated as an array in L31! what gives? ... oh wait, there's a transmutation in L26, so that line invalidates the docs."). compilers also don't care _too much_ about this **in some cases** because they can just use another variable name internally (and they're pretty clever at that: if they can prove that all codepaths leading to a certain use necessarily pass by the line where the underlying type is being changed, they do treat it as the new type!), but in some other cases, they have to give up and say "well, this can be either a string or an array. we trust the developer that they know what they are doing, but will crash later otherwise".

if, on the other hand, a new name is explicitly used to refer to the arrayified string, both humans and compilers have a much better time comprehending / optimizing the code. bonus points if the new name conveys the fact that it has been arrayified. also, if this program were to be written in [TypeScript](http://typescriptlang.org/) (highly recommended to write code in TypeScript _instead of_ JavaScript. the amount of bugs and regressions it has saved my team is astounding).

avc278: Once I'm further into the book around Sept/Oct, I'll slowly migrate over to TS. I don't know much about it yet, but good to know that it helps prevent these bad practices with re-assigning variables between types.

lhchavez: that should probably give us enough time to set the unit tests up for TS!

PR #59, opened 2020-08-01
avc278 - 1.04 - Javascript

avc278:
Added my solution to the Chapter 1 Problem 4 folder.

### DIFF
--- a/JavaScript/chapter01/p04_palinperm/avc278.js
+++ b/JavaScript/chapter01/p04_palinperm/avc278.js
@@ -1,0 +1,54 @@
+// Palindrome Permutation: Given a string, write a function to check if it is a permutation of a palindrome. A
+// palindrome is a word or phrase that is the same forwards and backwards. A permutation is a rearrangement of letters.
+// The palindrome does not need to be limited to just dictionary words.
+
+const assert = require("assert");
+
+/**
+ * Checks if the input string can be rearranged into a palindrome
+ * @param  {string} str input string to check against
+ * @return {boolean}    whether the input string can be rearranged into a palindrome
+ *
+ * We don't actually have to rearrange the string to solve this problem; we only need to check its contents.
+ * If the string is of odd  length, then each character in the string has to appear an even number of times, except one.
+ * If the string is of even length, then each character in the string has to appear an even number of times.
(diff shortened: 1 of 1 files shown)


JavaScript/chapter01/p04_palinperm/avc278.js
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * Runtime: O(N)
+ * Space:   O(N)
+ *
+ */
+const palindromePerm = (str) => {
+  const letterFreq = {};
+  let length = 0;
+  for (const letter of str) {
+    if (letter !== " ") {
```
lhchavez, line 26: small optional readability suggestion: here (and in L37, L38), the [guard clause pattern](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) can be used to reduce nesting and improve readability:

```suggestion
    if (letter === " ") continue;  // or in the next line.
```

JavaScript/chapter01/p04_palinperm/avc278.js
```diff
@@ -0,0 +1,54 @@
...
+        (letterFreq[letter.toLowerCase()] || 0) + 1;
+      length += 1;
+    }
+  }
+
+  const keys = Object.keys(letterFreq);
+
+  let slack = length % 2 == 1;
+  for (const key of keys) {
+    if (letterFreq[key] % 2 !== 0) {
```
lhchavez, line 37: given that `key` is only used to get the value of the key/value pair, it's probably better to express that more directly with [`Object.values()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Object/values):

```suggestion
  for (const letterFreq of Object.values(letterFreqs)) {  // assuming L23 is renamed to be plural.
    if (letterFreq % 2 === 0) continue;
```

JavaScript/chapter01/p04_palinperm/avc278.js
avc278: oops... I actually knew this one :P

avc278: Completely forgot to update this one. Should be fixed now

PR #61, opened 2020-08-02
avc278 - 1.06 - Javascript

avc278:
Added my solution to the Chapter 1 Problem 6 folder.

### DIFF
--- a/JavaScript/chapter01/p06_string_compression/avc278.js
+++ b/JavaScript/chapter01/p06_string_compression/avc278.js
@@ -10,8 +10,13 @@ const assert = require("assert");
  * @param  {string}  str input string to compress
  * @return {string}      either the compressed string, or the string
  *
- * Runtime: O()
- * Space:   O()
+ * Storing the outputStr is at worst size 2N for an input string whose size is N, which is still O(N) additional space.
+ * Storing and updating the pointers `idx`, `next`, and `freq` take O(1) additional space and O(1) time.
+ * As we iterate through the str in the outer while loop, in one extreme, assuming no repeating letters in a sequence,
+ * idx increments by 1 each iteration, and so it's O(N) runtime. On the other extreme, assuming all repeating letters,
+ * this would be O(N) on the outer loop, and O(N) in the inner while loop.
+ * Runtime: O(N^2)
+ * Space:   O(N)
  *
  */
 const compress = (str) => {
--- a/JavaScript/chapter01/p06_string_compression/avc278.js
+++ b/JavaScript/chapter01/p06_string_compression/avc278.js
@@ -10,35 +10,39 @@ const assert = require("assert");
  * @param  {string}  str input string to compress
  * @return {string}      either the compressed string, or the string
  *
- * Storing the outputStr is at worst size 2N for an input string whose size is N, which is still O(N) additional space.
+ * Storing the outputStrArr is at worst size 2N for an input string whose size is N (still O(N) additional space).
  * Storing and updating the pointers `idx`, `next`, and `freq` take O(1) additional space and O(1) time.
  * As we iterate through the str in the outer while loop, in one extreme, assuming no repeating letters in a sequence,
  * idx increments by 1 each iteration, and so it's O(N) runtime. On the other extreme, assuming all repeating letters,
- * this would be O(N) on the outer loop, and O(N) in the inner while loop.
- * Runtime: O(N^2)
+ * this would be O(N) on the outer loop, and O(N) in the inner while loop. The trick to knowing this is still O(N)
+ * and not O(N^2) is the fact that we don't visit any one index pointer more than once in our two while loops.
+ * Both loops are still just part of the same loop, which results in O(N).
+ * 
+ * Runtime: O(N)
  * Space:   O(N)
  *
  */
 const compress = (str) => {
-  let outputStr = "";
+  let outputStrArr = [];
   let idx = 0;
   let next = 1;
+
   while (idx < str.length) {
     const currLetter = str[idx];
     let freq = 1;
-    while (true) {
-      if (currLetter === str[next]) {
-        next += 1;
-        freq += 1;
-      } else {
-        outputStr += `${currLetter}${freq}`;
-        idx += freq;
-        next = idx + 1;
-        break;
-      }
+
+    while (next < str.length && currLetter === str[next]) {
+      next += 1;
+      freq += 1;
     }
+
+    outputStrArr.push(currLetter);
+    outputStrArr.push(freq);
+    idx += freq;
+    next = idx + 1;
   }
-  return outputStr.length < str.length ? outputStr : str;
+
+  return outputStrArr.length < str.length ? outputStrArr.join("") : str;
 };
 
 describe(module.filename, () => {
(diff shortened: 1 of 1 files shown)


JavaScript/chapter01/p06_string_compression/avc278.js
```diff
@@ -0,0 +1,54 @@
...
+ * Returns the compressed version of the string if applicable
+ * @param  {string}  str input string to compress
+ * @return {string}      either the compressed string, or the string
+ *
+ * Storing the outputStr is at worst size 2N for an input string whose size is N, which is still O(N) additional space.
+ * Storing and updating the pointers `idx`, `next`, and `freq` take O(1) additional space and O(1) time.
+ * As we iterate through the str in the outer while loop, in one extreme, assuming no repeating letters in a sequence,
+ * idx increments by 1 each iteration, and so it's O(N) runtime. On the other extreme, assuming all repeating letters,
+ * this would be O(N) on the outer loop, and O(N) in the inner while loop.
+ * Runtime: O(N^2)
```
lhchavez, line 18: IIUC the statement in L17 is true. yet this runtime analysis is not correct. how can that be? what is the correct runtime?

(although in a weird twist of things, it turns out that the implementation as written _is_ O(N²) due to the fact that `outputStr` is being concatenated character by character. JavaScript strings being immutable causes a new string of length O(N) to be allocated in each of the O(N) rounds of the external loop, which causes the whole thing to be quadratic. the solution is to have an array of strings and join them all at the end: arrays do still need to be reallocated when they are appended, _but_ they are created with some additional capacity to avoid having to reallocate every single time a new element is appended).

JavaScript/chapter01/p06_string_compression/avc278.js
```diff
@@ -0,0 +1,54 @@
...
+ */
+const compress = (str) => {
+  let outputStr = "";
+  let idx = 0;
+  let next = 1;
+  while (idx < str.length) {
+    const currLetter = str[idx];
+    let freq = 1;
+    while (true) {
+      if (currLetter === str[next]) {
```
lhchavez, line 30: this is accessing the array out-of-bounds. JavaScript is lax enough that it just returns `undefined` when that happens, but other languages are not as lenient.

JavaScript/chapter01/p06_string_compression/avc278.js
```diff
@@ -0,0 +1,54 @@
...
+ *
+ */
+const compress = (str) => {
+  let outputStr = "";
+  let idx = 0;
+  let next = 1;
+  while (idx < str.length) {
+    const currLetter = str[idx];
+    let freq = 1;
+    while (true) {
```
lhchavez, line 29: this can be replaced with a `for` loop or another `while` with a condition.

JavaScript/chapter01/p06_string_compression/avc278.js
avc278: Hmm, so let me walk you through my thought process for this as it was a bit confusing, and I'm still not sure if I fully understand it:

Say you have a string "aaaaa", which gets compressed to "a5".
Let's step through the algorithm
`outputStr = '', idx = 0, next = 1`
enter while loop since idx (0) is less than str.length (5)
`currLetter = 'aaaaa'[0] = 'a', freq = 1`
enter true while loop
`'a' === 'aaaaa'[1] (true)`
so we increment next and freq `next=2, freq=2`
`'a' === 'aaaaa'[2] (true)`
so we increment next and freq `next=3, freq=3`
`'a' === 'aaaaa'[3] (true)`
so we increment next and freq `next=4, freq=4`
`'a' === 'aaaaa'[4] (true)`
so we increment next and freq `next=5, freq=5`
`'a' === 'aaaaa'[5] (false, this is actually a bug that you pointed out down below which I'll fix, but something that JS is so nice to take of for us)`
so we hit the `else` and create our string `outputStr = '' + currLetter(a)freq(5)` => 'a5'
and update idx from 0 to 0+5 = 5, and next from 5 to 6.
We go back to the start of our while loop and break since 5 !< 5.

In the inner loop for this case, we iterated 5 times (the length of the string), and in the outer loop we only iterate once. So I guess for this case, the run time was actually O(N)?

=========

Perhaps for a different example "aabb"
`outputStr = '', idx = 0, next = 1`
enter outer while loop
`currLetter = 'aabb'[0] = 'a', freq = 1`
enter true while loop
`'a' === 'aabb'[1] (true)`
so we increment next and freq `next=2, freq=2`
`'a' === 'aabb'[2] (false)
so we hit the `else` and update our string `outputStr = '' + currLetter(a)freq(2)` => `a2`
and update idx from 0 to 0+2 = 2, and next from 2 to 3.
So this is O(1) on the outer loop, but O(a) where a is the length of the substring containing connected a's.

`currLetter = 'aabb'[2] = 'b', freq = 1`
enter true while loop
`'b' === 'aabb'[3] (true)`
so we increment next and freq `next=4, freq=2`
`'b' === 'aabb'[4] (false)`
so we hit the `else` and update our string `outputStr = 'a2' + currLetter(b)freq(2)` => `a2b2`
and update idx from 2 to 2+2 = 4, and next from 4 to 5.
Again this is the second iteration on the outer loop and so I think O(d) where d is the number of distinct elements side by side, and inner loop O(b) where b is the length of the substring containing connected b's.

Overall runtime for this example was O(d * (a + b))?

lhchavez: the correct runtime complexity is O(N). Normally we multiply the complexities of nested loops, _but_ in cases like this we can use a different strategy: count the number of times the `str[idx]` and `str[next]` expressions are going to be evaluated. every iteration of the inner loop it will evaluate `str[next]` exactly once (with a different `next` value every time), and every iteration of the outer loop it will evaluate `str[idx]` exactly once (with a different `idx` value every time). This means that effectively both loops are really just one loop being interleaved, which yields a final answer of O(N).

avc278: >both loops are really just one loop being interleaved

This is exactly what I wasn't getting at first, but I understand now. Thanks! 👏

JavaScript/chapter01/p06_string_compression/avc278.js
```diff
@@ -10,35 +10,39 @@ const assert = require("assert");
...
+
+    while (next < str.length && currLetter === str[next]) {
+      next += 1;
+      freq += 1;
     }
+
+    outputStrArr.push(currLetter);
+    outputStrArr.push(freq);
+    idx += freq;
+    next = idx + 1;
```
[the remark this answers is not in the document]
avc278: yeah for sure

JavaScript/chapter01/p06_string_compression/avc278.js
```diff
@@ -10,35 +10,39 @@ const assert = require("assert");
...
-        outputStr += `${currLetter}${freq}`;
-        idx += freq;
-        next = idx + 1;
-        break;
-      }
+
+    while (next < str.length && currLetter === str[next]) {
+      next += 1;
+      freq += 1;
     }
```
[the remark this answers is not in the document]
avc278: Hmm yeah that's definitely a good point to remember. In my head I like updating next after updating idx since that makes the most sense to me, but I see why it's cleaner and beneficial to update it before the inner while loop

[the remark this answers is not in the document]
avc278: Oh yeah one last thing:

I see that unary increment suggestion, which is totally valid, but I prefer following airbnb's style guide on this whenever I can remember to do so

https://github.com/airbnb/javascript#variables--unary-increment-decrement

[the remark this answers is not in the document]
lhchavez: fair enough. i'm attuned to [Google's style guide](https://google.github.io/styleguide/jsguide.html) (it makes no mention of the unary operators, but in practice most folks use them) so please excuse any Google-isms.

As long as you follow _a_ style guide, everything is good!

PR #62, opened 2020-08-02
avc278 - 1.07 - Javascript

avc278:
Added my solution to the Chapter 1 Problem 7 folder.
A bit unsure on the docstring even after searching stack overflow
` * @param  {Array.<number[]>} matrix input 2d array to rotate`
That's the best I can come up with aside from number[][]

btw, I'll have to rebase once 1.6 is merged in since I accidentally branched off 1.6 instead of master for this one.

### DIFF
--- a/JavaScript/chapter01/p07_rotate_matrix/avc278.js
+++ b/JavaScript/chapter01/p07_rotate_matrix/avc278.js
@@ -68,6 +68,20 @@ describe(module.filename, () => {
       [1, 3],
     ]);
   });
+  it("should rotate a 3x3 matrix 90 degrees", () => {
+    const matrix = [
+      [1, 2, 3],
+      [4, 5, 6],
+      [7, 8, 9],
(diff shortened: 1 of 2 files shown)


JavaScript/chapter01/p07_rotate_matrix/avc278.js
```diff
@@ -0,0 +1,104 @@
...
+    rotateMatrix(matrix);
+    assert.deepEqual(matrix, [
+      [14, 18, 22, 26],
+      [13, 17, 21, 25],
+      [12, 16, 20, 24],
+      [11, 15, 19, 23],
+    ]);
+  });
+  it("should bring the matrix to its original position after rotating 4 times.", () => {
+    const matrix = [
```
lhchavez, line 88: it may be a good idea to have a 3x3 or 5x5 test case where it is only rotated once.

PR #63, opened 2020-08-04
avc278 - 1.08 - Javascript

avc278:
Added my solution to the Chapter 1 Problem 8 folder.

### DIFF
--- a/JavaScript/chapter01/p08_zero_matrix/avc278.js
+++ b/JavaScript/chapter01/p08_zero_matrix/avc278.js
@@ -1,0 +1,94 @@
+// Zero Matrix: Write an algorithm such that if an element in an MxN matrix is 0, its entire row and columns are set to 0.
+
+const assert = require("assert");
+
+/**
+ * Overwrite rows and columns who have at least 1 zero in them with zeros.
+ * @param  {Array.<number[]>} matrix input 2d array to mutate
+ * @return {null}
+ *
+ * Okay, so I have a jank solution that assumes no element in the array is currently -999.
+ * We iterate through the array, and when we find a 0, we flag its row and column by updating all elements in there to
+ * -999. The outer two for loop iterations is O(M * N) time, and each inner for loop is O(M) and O(N), respectively,
+ * resulting in a grand total of O((M * N) * (M + N)) time complexity, while updating takes O(1) additional space.
+ * When we take a second pass at the matrix, we update all elements that have been flagged -999 to 0, which is O(N^2)
+ * time and O(1) additional space.
+ *
+ * Runtime: O((M * N) * (M + N))
+ * Space:   O(1)
+ *
+ */
+const zeroMatrix = (arr) => {
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[0].length; j++) {
+      if (arr[i][j] !== 0) continue;
+
+      for (let k = 0; k < arr.length; k++) {
+        if (arr[k][j] !== 0) arr[k][j] = -999;
+      }
+
+      for (let l = 0; l < arr[0].length; l++) {
+        if (arr[i][l] !== 0) arr[i][l] = -999;
+      }
+    }
+  }
+
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[0].length; j++) {
+      if (arr[i][j] === -999) {
+        arr[i][j] = 0;
+      }
+    }
+  }
+};
+
+describe(module.filename, () => {
+  it("should not affect a matrix of size 0", () => {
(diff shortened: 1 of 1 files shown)


JavaScript/chapter01/p08_zero_matrix/avc278.js
```diff
@@ -0,0 +1,94 @@
+// Zero Matrix: Write an algorithm such that if an element in an MxN matrix is 0, its entire row and columns are set to 0.
+
+const assert = require("assert");
+
+/**
+ * Overwrite rows and columns who have at least 1 zero in them with zeros.
+ * @param  {Array.<number[]>} matrix input 2d array to mutate
+ * @return {null}
+ *
+ * Okay, so I have a jank solution that assumes no element in the array is currently -999.
```
lhchavez, line 10: haha, i laughed out loud when i read this.

ahem. would it be possible to replace the "magic number" -999 by something that cannot be confused with a valid number? say... [`NaN`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN)? it is still represented as a [`Number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) (so it does not have the undesirable property of `undefined` or `null` that the type must be relaxed to be able to express this constant).

normally whenever there's a need to have a sentinel value, it should be chosen in a way that it is not representable as a valid item in the range of the type. for floating-point numbers, `NaN` is a good candidate, but typically `null` or `undefined` should be preferred.

JavaScript/chapter01/p08_zero_matrix/avc278.js
```diff
@@ -0,0 +1,94 @@
...
+ * We iterate through the array, and when we find a 0, we flag its row and column by updating all elements in there to
+ * -999. The outer two for loop iterations is O(M * N) time, and each inner for loop is O(M) and O(N), respectively,
+ * resulting in a grand total of O((M * N) * (M + N)) time complexity, while updating takes O(1) additional space.
+ * When we take a second pass at the matrix, we update all elements that have been flagged -999 to 0, which is O(N^2)
+ * time and O(1) additional space.
+ *
+ * Runtime: O((M * N) * (M + N))
+ * Space:   O(1)
```
lhchavez, line 18: can you think of a runtime/space tradeoff that can run in O(M * N) time?

JavaScript/chapter01/p08_zero_matrix/avc278.js
avc278: That makes a lot of sense. At work, my boss and I agreed to hard code years to only be four places because "once it's between the year 9999 and 10000, it's no longer something we have to worry about 😉 "

JavaScript/chapter01/p08_zero_matrix/avc278.js
avc278: Yeah, the only other option I can think of would be to make an additional two arrays of bools O(M + N) set to false. Each time i, j in the original array is found to be zero, add i to one array, and j to the other. Once we're done iterating through the entire matrix, we can iterate through the first generated array of row bools and update all matrix elements in that row to 0s, then do the same to the column bools.
Runtime: O(M*N)
Space: O(M + N)

lhchavez: yup, that works!

JavaScript/chapter01/p08_zero_matrix/avc278.js
```diff
@@ -7,37 +7,46 @@ const assert = require("assert");
...
+  if (arr.length === 0 || arr[0].length === 0) return;
+  const rows = [];
+  const cols = [];
+
+  for (let i = 0; i < arr.length; i++) {
+    rows.push(false);
+  }
+  for (let i = 0; i < arr[0].length; i++) {
+    cols.push(false);
+  }
```
lhchavez, line 29: Using [`Array.fill()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill) makes a bit more concise code:

```suggestion
  const rows = new Array(arr.length).fill(false);
  const cols = new Array(arr[0].length).fill(false);
```

avc278: I'm a bit embarrassed to say I've never seen `Array.fill()` up until this point in my journey with JS. But yeah, you're totally right. This cleans it up by a lot.

lhchavez: don't be embarrassed! https://xkcd.com/1053/

nobody is expected to know _the whole JavaScript API_. so all's good!

PR #64, opened 2020-08-04
avc278 - 1.09 - Javascript

avc278:
Added my solution to the Chapter 1 Problem 9 folder.
This is the final problem in chapter 1 🎉

### DIFF
--- a/JavaScript/chapter01/p09_string_rotation/avc278.js
+++ b/JavaScript/chapter01/p09_string_rotation/avc278.js
@@ -27,11 +27,6 @@ describe(module.filename, () => {
     const s2 = "hello";
     assert.ok(!isSubstring(s1, s2));
   });
-  it("should return false when s2 is a valid rotation of s1, but not of the same length, hence not a rotation", () => {
-    const s1 = "wat";
-    const s2 = "erbottlewat";
-    assert.ok(!isSubstring(s1, s2));
-  });
   it("should return true when s2 is a valid rotation of s1", () => {
     const s1 = "waterbottle";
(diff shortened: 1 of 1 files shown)


JavaScript/chapter01/p09_string_rotation/avc278.js
```diff
@@ -0,0 +1,40 @@
...
+  return `${s1}${s1}`.includes(s2);
+};
+
+describe(module.filename, () => {
+  it("should return false when both input strings are not of the same length", () => {
+    const s1 = "hi";
+    const s2 = "hello";
+    assert.ok(!isSubstring(s1, s2));
+  });
+  it("should return false when s2 is a valid rotation of s1, but not of the same length, hence not a rotation", () => {
```
lhchavez, line 30: this statement is a bit confusing.

avc278: Yeah, honestly it was a dumb test case to begin with since we never even reach there in the first place, unless I had my solution as a one liner like
```
return `${s1}${s1}`.includes(s2) && s1.length === s2.length;
```

Only then would this have been a valid test case in my opinion.

lhchavez: oh i meant that the statement claims that s2 is a valid rotation.... but later it claims that it's not.

that works in proofs by contradiction, but in prose it needs some refactoring to avoid confusing the reader. so "should return false when s2 is a substring of s1, but since they're not the same length it is not a rotation" could work. in any case, this is ready to land!

PR #65, opened 2020-08-09
avc278 - 2.01 - Javascript

avc278:


JavaScript/chapter02/p01_remove_dups/avc278.js
```diff
@@ -0,0 +1,101 @@
+// Remove Dups: Write code to remove duplicates from an unsorted linked list.
+// Follow Up: How would you solve this problem if a temporary buffer is not allowed?
+
+const assert = require("assert");
+
+class LinkedListNode {
+  constructor(val, next) {
+    this.val = val === undefined ? 0 : val;
```
lhchavez, line 8: it's probably better to leave the `val` unchanged: normally containers are thought to be generic data structures that can store anything without imposing any additional restrictions.

in other languages, any kind of restrictions on the values are handled by the type system (by making the containers generic and then enforcing that the values are of a certain type), but JavaScript does not have this luxury :(

JavaScript/chapter02/p01_remove_dups/avc278.js
```diff
@@ -0,0 +1,101 @@
...
+ * @return {null}
+ *
+ * Keep a temporary set for storing values we pass through, of maximum size N where N is the number of nodes in the
+ * input linked list in the worst case. This requires an additional O(N) space. We store two pointers as we traverse
+ * through the linked list, skipping nodes with values already in our set.
+ * Runtime: O(N)
+ * Space:   O(N)
+ *
+ */
+const removeDups1 = (LL) => {
```
lhchavez, line 25: per https://github.com/airbnb/javascript#naming--camelCase, the name `LL` might be confused with a class. how about `list`? same in L56.

JavaScript/chapter02/p01_remove_dups/avc278.js
```diff
@@ -0,0 +1,101 @@
...
+
+describe(module.filename, () => {
+  it("should return the linked list without duplicates", () => {
+    const ll6 = new LinkedListNode(4);
+    const ll5 = new LinkedListNode(2, ll6);
+    const ll4 = new LinkedListNode(4, ll5);
+    const ll3 = new LinkedListNode(3, ll4);
+    const ll2 = new LinkedListNode(2, ll3);
+    let ll1 = new LinkedListNode(2, ll2);
+    const head = ll1;
```
lhchavez, line 78: it might be a good idea to create a function that takes a list and creates a linked list out of it. it'll be useful to copy+paste for the rest of the chapter.

JavaScript/chapter02/p01_remove_dups/avc278.js
```diff
@@ -0,0 +1,101 @@
...
+    removeDups1(ll1);
+    while (ll1 !== null && expectedLL1 !== null) {
+      assert.equal(ll1.val, expectedLL1.val);
+      ll1 = ll1.next;
+      expectedLL1 = expectedLL1.next;
+    }
+
+    ll1 = head;
+    expectedLL1 = expectedHead;
+    removeDups2(ll1);
```
lhchavez, line 94: would it be a good idea to create a parameterized test (where the parameter is each implementation) for this so that each test case follows the [Arrange, Act, Assert](http://wiki.c2.com/?ArrangeActAssert) pattern AND avoids repetition?

JavaScript/chapter02/p01_remove_dups/avc278.js
```diff
@@ -0,0 +1,101 @@
...
+    const expectedLL2 = new LinkedListNode(3, expectedLL3);
+    let expectedLL1 = new LinkedListNode(2, expectedLL2);
+    const expectedHead = expectedLL1;
+
+    removeDups1(ll1);
+    while (ll1 !== null && expectedLL1 !== null) {
+      assert.equal(ll1.val, expectedLL1.val);
+      ll1 = ll1.next;
+      expectedLL1 = expectedLL1.next;
+    }
```
lhchavez, line 90: similar to above, it might be a good idea to create a function that compares two linked lists.

also, can you think of a way in which the algorithm could be doing something wrong **and still** pass this test?

JavaScript/chapter02/p01_remove_dups/avc278.js
```diff
@@ -0,0 +1,101 @@
...
+ * Runtime: O(N)
+ * Space:   O(N)
+ *
+ */
+const removeDups1 = (LL) => {
+  const seen = new Set();
+  let currNode = LL;
+  let next = LL.next;
+  seen.add(currNode.val);
+  while (next !== null) {
```
lhchavez, line 30: _maybe_ this can be expressed as a `for` loop:

```javascript
for (let nextNode = currNode.next; nextNode.next !== null; nextNode = nextNode.next) {
```

same in L60 (and maybe also L58)

avc278: Hmm yeah that works.

I'm having a hard time, though, seeing what the benefit of doing it that way is, as opposed to just using a regular while loop.
Is the purpose here to clean up the variable initialization, and increment-related code, so as to only have the duplication-related code in the loop?

lhchavez: yup! programming is all about building using patterns (where they make sense). by writing the loop in this way, it's more clear that a "normal" traversal is being done instead of having to read a bit more of the code to realize that same information.

of course sometimes coercing the code to fit a pattern is detrimental, so it's not a universally-applicable rule :/

if this were production code, making a `LinkedList` that implements the [iterator protocol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols) might be even better so that it could use a normal [`for ... of`]( https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) loop. but here it may be way overkill.

JavaScript/chapter02/p01_remove_dups/avc278.js
avc278: I _think_ I accounted for it in the upcoming commit. Basically just adding a check for a null list at the start of the function. I noticed my tests were failing when I checked the functions against an empty input.
`if (!list) return;`

JavaScript/chapter02/p01_remove_dups/avc278.js
```diff
@@ -0,0 +1,110 @@
...
+
+const arrayToLinkedList = (arr) => {
+  let tail = null;
+  for (let i = arr.length - 1; i >= 0; i--) {
+    tail = new LinkedListNode(arr[i], tail);
+  }
+  return tail;
+};
+
+const compareLinkedLists = (A, B) => {
```
lhchavez, line 21: what should the following expression evaluate to?

```javascript
compareLinkedLists(arrayToLinkedList([]), arrayToLinkedList(['hi!']))
```

JavaScript/chapter02/p01_remove_dups/avc278.js
```diff
@@ -0,0 +1,110 @@
+// Remove Dups: Write code to remove duplicates from an unsorted linked list.
+// Follow Up: How would you solve this problem if a temporary buffer is not allowed?
+
+const assert = require("assert");
+
+class LinkedListNode {
+  constructor(val, next) {
+    this.val = val === undefined ? null : val;
```
lhchavez, line 8: optional: there's a [snazzy new way](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) to express this!

```suggestion
    this.val = val ?? null;
```

avc278: This is definitely going to take some time to get used to.
But it's definitely superior for code gold.

JavaScript/chapter02/p01_remove_dups/avc278.js
avc278: Oh interesting. That didn't even occur to me to try initially. Good catch.
This will definitely be fixed in the next commit.

JavaScript/chapter02/p01_remove_dups/avc278.js
```diff
@@ -0,0 +1,117 @@
...
+
+const arrayToLinkedList = (arr) => {
+  let tail = null;
+  for (let i = arr.length - 1; i >= 0; i--) {
+    tail = new LinkedListNode(arr[i], tail);
+  }
+  return tail;
+};
+
+const compareLinkedLists = (A, B) => {
```
lhchavez, line 21: what about this one?

```javascript
compareLinkedLists(arrayToLinkedList([1]), arrayToLinkedList([1, 2]))
```

L23 is the solution, but it's in the wrong place.

JavaScript/chapter02/p01_remove_dups/avc278.js
```diff
@@ -20,14 +20,14 @@ const arrayToLinkedList = (arr) => {
...
-  if (!A || !B) return false;
 
   while (A !== null && B !== null) {
     if (A.val !== B.val) return false;
     A = A.next;
     B = B.next;
   }
-  return true;
+
+  return !((A && !B) || (!A && B));
```
lhchavez, line 30: this could be simplified through [de Morgan's laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws):

```javascript
   !((A && !B) || (!A && B))
=> !(A && !B) && !(!A && B)
=> (!A || B) && (A || !B)
```

Apply a bit of [Distributivity](https://en.wikipedia.org/wiki/Distributive_property):

```javascript
   (!A || B) && (A || !B)
=> ((!A || B) && A) || ((!A || B) && !B)
=> (A && (!A || B)) || (!B && (!A || B))
=> (A && !A || A && B) || (!B && !A || !B && B)
=> (A && B) || (!B && !A)
```

This final one is another way of expressing the [Logical biconditional](https://en.wikipedia.org/wiki/Logical_biconditional):

```javascript
   (A && B) || (!B && !A)
=> (A !== null) === (B !== null)
```

based on the fact that the code is no longer in the loop in L24-L28, it means that it _has_ to be the case that `!(A !== null && B !== null) => A !== null || B !== null`. This means that the only way in which the previous expression can be true is

```javascript
return (A !== null) && (B !== null);
```

which is a more concise way of saying:

```javascript
if (!((A !== null) && (B !== null))) return false;
return true;
```

which in turn takes us back to:

```javascript
if (!A || !B) return false;
return true;
```

which is the reason why I mentioned that L23 was the solution! (although in prouction code `return (A !== null) && (B !== null);` may be preferred since it's simpler and the intention of "return true if BOTH linked lists have finished" is easier to grasp by humans).

(how did i get to all this result? i knew the answer upfront, so i cheated a bit since I was able to work my way backwards from the expected answer, BUT if i had not known the answer, one technique is to evaluate the expression with all possible boolean values and print out the logic table. look that table up in https://en.wikipedia.org/wiki/Truth_table#Binary_operations to see what's the immediate simplification. note that sometimes it's better to choose a suboptimal simplification if it allows the code flow to be easier to understand, especially if techniques like the [guard clause pattern](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) is used).

avc278: Very interesting. Thank you for the in-depth explanation

PR #67, opened 2020-08-15
avc278 - 2.03 - Javascript

avc278:


JavaScript/chapter02/p03_delete_middle_node/avc278.js
```diff
@@ -0,0 +1,66 @@
...
+
+const compareLinkedLists = (A, B) => {
+  if (!A && !B) return true;
+
+  while (A !== null && B !== null) {
+    if (A.val !== B.val) return false;
+    A = A.next;
+    B = B.next;
+  }
+  return !((A && !B) || (!A && B));
```
lhchavez, line 32:
```suggestion
  return A !== null && B !== nulll;
```

avc278: Reading back on the explanation in #65, I do think we want to return true when we have exhausted A and B.
So `return A === null && B === null` would be valid here, or just `return !A && !B`.

lhchavez: err... yes, i messed up the negations ^^;;

avc278: Oh btw, while we're here, would you mind if I made a separate directory in chapter 2 for re-usable functions/classes just to DRY things out?
Something like `JavaScript/chapter02/utils/avc278.js`

lhchavez: creating a directory to host re-usable functions/classes sounds good!

in some circles, the name "utils" is frowned upon (e.g. https://softwareengineering.stackexchange.com/a/175190, https://www.vojtechruzicka.com/avoid-utility-classes/), so `JavaScript/lib/avc278/linkedlist.js` may be a better location.

avc278: Oo sounds good to me. Keep an eye out for it in my next PR.
I'll also sneak in refactors to the previous solutions in chapter 2.

PR #68, opened 2020-08-16
Refactor: JavaScript/lib

avc278:
Creates a new directory `lib` within the `JavaScript` solutions directory to store re-usable functions and classes.

The general pattern to follow would be: `JavaScript/lib/{{ github_handle }}/{{ category }}`

So for the case of "Chapter 2: Linked Lists", I would add a new file `JavaScript/lib/avc278/linkedlist.js`.


JavaScript/lib/avc278/linkedlist.js
```diff
@@ -0,0 +1,43 @@
...
+const compareLinkedLists = (A, B) => {
+  if (!A && !B) return true;
+
+  while (A !== null && B !== null) {
+    if (A.val !== B.val) return false;
+    A = A.next;
+    B = B.next;
+  }
+
+  return !((A && !B) || (!A && B));
```
lhchavez, line 27:
```suggestion
  return !A && !B;
```

avc278: omg...
that's what I get for blindly c/p

PR #71, opened 2020-08-16
avc278 - 2.04 - Javascript

avc278:


JavaScript/chapter02/p04_partition/avc278.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+    head = head.next;
+  }
+  less.next = moreHead;
+  more.next = null;
+  return lessHead;
+};
+
+describe(module.filename, () => {
+  it("should return the partitioned linked list when the partition value exists in the linked list", () => {
```
lhchavez, line 61: what should the answer to

```javascript
partition(arrayToLinkedList([10]), 5);
```

be?

JavaScript/chapter02/p04_partition/avc278.js
```diff
@@ -0,0 +1,81 @@
...
+ * @param   {LinkedListNode} head input linked list to be partitioned
+ * @return  {LinkedListNode}      partitioned linked list
+ *
+ * For this problem, we need to store two linked lists: one for values lesser than `partitionValue`, and one for values
+ * greater than `partitionValue`. We also need to store pointers for the heads of these linked lists as we need to merge
+ * them at the end. Iterating once through the original linked list makes it so the runtime is O(N). We connect nodes
+ * together in the less and more lists until we exhaust the original list. At this point, we connect the end of less to
+ * the head of more, set `more`'s next to null and return the head of less.
+ * Runtime: O(N)
+ * Space:   O(N)
```
lhchavez, line 26: would it be possible to elaborate on the O(N) space requirement?

JavaScript/chapter02/p04_partition/avc278.js
```diff
@@ -0,0 +1,81 @@
...
+// Output: 3 -> 1 -> 2 -> 10 -> 5 -> 5 -> 8
+
+const assert = require("assert");
+const {
+  arrayToLinkedList,
+  compareLinkedLists,
+} = require("../../lib/avc278/linkedlist");
+
+/**
+ * Partitions a linked list based on the input partition, with lesser values to the left and greater values to the right
```
lhchavez, line 16: JavaScript does not have a way to express whether a function will mutate the arguments or not (i.e. if it is considered to be [destructive](https://www.cs.cmu.edu/~112/notes/notes-1d-lists.html#destructiveFunctions)), so that needs to be documented.

non-destructive functions typically have slightly different requirements, since they need to make a copy of the input list (which translates to them typically requiring more memory and being a bit slower). therefore, they are not necessarily better (or worse) than their destructive counterparts.

JavaScript/chapter02/p04_partition/avc278.js
```diff
@@ -0,0 +1,81 @@
...
+ * greater than `partitionValue`. We also need to store pointers for the heads of these linked lists as we need to merge
+ * them at the end. Iterating once through the original linked list makes it so the runtime is O(N). We connect nodes
+ * together in the less and more lists until we exhaust the original list. At this point, we connect the end of less to
+ * the head of more, set `more`'s next to null and return the head of less.
+ * Runtime: O(N)
+ * Space:   O(N)
+ *
+ */
+const partition = (head, partitionValue) => {
+  let less;
```
lhchavez, line 30: nit: maybe call this `lessTail` (and `moreTail`).

typically whenever there is a distinction between two variables, it is preferred to name the two variants with a similar level of verbosity for a) making it easier to identify that they are related through analogy, b) highlighting the distinguishing factor a bit more, and c) allowing readers to identify whether the two variants are mutually exclusive or there may be some overlap between them.

JavaScript/chapter02/p04_partition/avc278.js
avc278: amateur hour over here 🤦

JavaScript/chapter02/p04_partition/avc278.js
avc278: Hmm, so looking at it again, we're never creating new linked list nodes, and so we don't allocate any new memory, so this is actually O(1). All we're doing is storing two node relationships at a time and two starting nodes.

JavaScript/chapter02/p04_partition/avc278.js
```diff
@@ -13,69 +13,98 @@ const {
...
       }
     }
     head = head.next;
   }
-  less.next = moreHead;
-  more.next = null;
-  return lessHead;
+
+  if (!lessHead && moreHead) {
+    head = moreHead;
```
lhchavez, line 60: note that JavaScript is a [pass by value](https://medium.com/nodesimplified/javascript-pass-by-value-and-pass-by-reference-in-javascript-fcf10305aa9c) language (like most managed languages. there are some exceptions, like PHP and C# where you can specify [how the parameters are passed by](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/ref#passing-an-argument-by-reference)), so L59-L63 are no-ops* and can be removed without changing the program's behavior: the effect of executing those lines won't be visible outside this function.

\* in compiler lingo, this is a case of [Dead code elimination](https://en.wikipedia.org/wiki/Dead_code_elimination): writes to variables that are guaranteed to never be read again are often removed by the compiler since there are no observable side effects of doing (or not doing) that computation. There are some cases in which the compiler will keep the code, like when the code is being compiled in non-production mode, when there is a chance that an exception is raised in the code (since that _will_ produce visible side-effects), or in interpreted languages when a debugger is attached.

JavaScript/chapter02/p04_partition/avc278.js
```diff
@@ -13,69 +13,98 @@ const {
...
   it("should return the partitioned linked list when the partition value exists in the linked list", () => {
     const arr = [3, 5, 8, 5, 10, 2, 1];
-    const ll1 = arrayToLinkedList(arr);
+    const ll = arrayToLinkedList(arr);
     const expectedArr = [3, 2, 1, 5, 8, 5, 10];
-    const expectedLL1 = arrayToLinkedList(expectedArr);
+    const expectedPartition = arrayToLinkedList(expectedArr);
 
-    partition(ll1, 5);
+    partition(ll, 5);
```
lhchavez, line 76: what should the effect of running

```javascript
const ll = arrayToLinkedList([10, 1]);
partition(ll, 5);
```

be?

JavaScript/chapter02/p04_partition/avc278.js
avc278: Hmm that makes sense. But if that's the case, then what I came up with initially was a hack to pass the two edge cases
- `partition(arrayToLinkedList[10], 5)`
- `partition(arrayToLinkedList[1], 5)`

In the meantime, I'll think of a way to compensate in the main section of the algorithm for these two scenarios if this is the case.

PR #73, opened 2020-08-27
avc278 - 2.06 - Javascript

avc278:


JavaScript/chapter02/p06_palindrome/avc278.js
```diff
@@ -0,0 +1,88 @@
...
+  while (fast !== null) {
+    fast = fast.next;
+    counter += 1;
+  }
+  counter /= 2;
+
+  while (counter > 0) {
+    slow = slow.next;
+    counter -= 1;
+  }
```
lhchavez, line 45: maybe

```suggestion
  for (let i = 0; i < counter / 2; i++) {
    slow = slow.next;
  }
```

since the number of times this loop needs to run is known beforehand.

avc278:
> code golf challenge: there is a way to combine the first two loops. That's just for style points, though, so let's merge this.

```
  while (fast !== null && fast.next !== null) {
    fast = fast.next.next;
    slow = slow.next;
  }
```
:V

PR #75, opened 2020-09-20
avc278 - 2.08 - Javascript

avc278:


JavaScript/chapter02/p08_loop_detection/avc278.js
```diff
@@ -0,0 +1,100 @@
...
+
+    if (slow !== fast) continue;
+    intersect = true;
+  }
+  slow = list;
+
+  while (slow !== fast) {
+    slow = slow.next;
+    fast = fast.next;
+  }
```
lhchavez, line 42: Can there be an explanation as to why this (and the loop in L30) work? Not that I don't believe the code is correct, but this is "magical" to most folks.

JavaScript/chapter02/p08_loop_detection/avc278.js
```diff
@@ -0,0 +1,100 @@
...
+ * Runtime: O(N)
+ * Space:   O(1)
+ *
+ */
+const loopDetection = (list) => {
+  let slow = list;
+  let fast = list;
+  let intersect = false;
+
+  while (!intersect) {
```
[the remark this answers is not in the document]
avc278: Oh wow, I haven't used a do while loop since college...
This definitely does clean things up though :)

JavaScript/chapter02/p08_loop_detection/avc278.js
avc278: Yeah, good point. I'll make a short write-up and update the solution file. This was definitely magic to me the first time I stumbled upon this solution a few weeks ago.

lhchavez (approved):
that's a _very_ nice explanation!

PR #77, opened 2020-10-30
Another O(N) solution using an Object

Ada-11:
Another solution without additional data structures but using objects, with performance O(N).

function isUnique3(str){
let obj = {}
for( let elem of str){
  if(obj.hasOwnProperty(elem)) obj[elem]++
  else obj[elem]=1
}
for( key in obj){
  if (obj[key]=== 1) return true
  else return false
}
} 

isUnique3('ada') // false
isUnique3("golden") // true


lhchavez:
dupe of #78

PR #78, opened 2020-10-30
Another solution for this exercise using Object

Ada-11:


JavaScript/chapter01/p01_is_unique/fvntr.js
```diff
@@ -14,3 +14,15 @@ describe(module.filename, () => {
...
+const isUnique2=(arr)=>{
+let obj = {}
+for (let elem of arr){
+  if(obj.hasOwnProperty(elem)) obj[elem]++
+  else obj[elem] = 1
+}
+
+return Object.keys(obj)
+}
+//isUnique2([1,1,1,2,2,2,2,3,3,3,3]) //returns[1,2,3]
```
lhchavez, line 28: can you create a new file (called `Ada-11.js`)? that way there's no way that multiple solutions conflict with each other.

also, it is considered a best practice (and thus strongly encouraged) to use unit tests to validate solutions: that way you can clearly see that a particular solution is passing (or failing) the tests quickly by executing it. We also have set up CI to automatically [run that every time](https://github.com/techqueria/data-structures-and-algorithms/runs/1333179854?check_suite_focus=true#step:4:5)!

you can see how to do that in L10-16 in this file or 
https://github.com/techqueria/data-structures-and-algorithms/blob/599ee5de6fcc405a6b9f1c2351197c26020ad7d4/JavaScript/chapter01/p01_is_unique/avc278.js#L50-L59

PR #79, opened 2020-12-23
Is unique branch

Ada-11:
Hi, 
This is a solution of this exercise using JS obejcts.

P.S. Am I supposed to run the mocha tests in my machine? 

THanks!

### DIFF
--- a/JavaScript/chapter01/p01_is_unique/ada11.js
+++ b/JavaScript/chapter01/p01_is_unique/ada11.js
@@ -1,0 +1,30 @@
+
+
+ /* Write a function isUnique that:
+Input: takes an array of integers
+Output: returns a deduped array of integers
+*/
+
+ //Solution using Object
+const isUnique = (arr)=>{
+ let obj = {}
+ for (let elem of arr){
+   if(obj.hasOwnProperty(elem)) obj[elem]++
+   else obj[elem] = 1
+ }
+
+ return Object.keys(obj)
+ }
+
+//test cases
+const assert = require('assert');
+  describe(module.filename, () => {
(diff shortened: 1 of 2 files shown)


JavaScript/chapter01/p01_is_unique/ada11.js
```diff
@@ -0,0 +1,30 @@
...
+
+ return Object.keys(obj)
+ }
+
+//test cases
+const assert = require('assert');
+  describe(module.filename, () => {
+    it("should return true on an input string with unique characters", () => {
+      assert.equal(isUnique1("tech"), true);
+      assert.equal(isUnique2("tech"), true);
```
lhchavez, line 24:
```suggestion
      assert.equal(isUnique("tech"), true);
```

there's only one implementation in this file. same below.

JavaScript/chapter01/p01_is_unique/avc278.js
```diff
@@ -1,3 +1,4 @@
+
```
lhchavez, line 1: this seems like an unintended change in another person's file:

```suggestion
```

lhchavez (commented):
> Hi,

welcome!

> This is a solution of this exercise using JS obejcts.
> 
> P.S. Am I supposed to run the mocha tests in my machine?

yes. that way you can quickly see if the algorithm as implemented solves the desired problem. in this case, the tests seem to not be passing, so it is likely that the solution is not correct :(

rather than returning an array, the function should return `true` if all characters in the string are unique or `false` otherwise (the name is a hint: `isUnique` is something that hints at a yes/no answer).

> 
> THanks!

PR #82, opened 2021-08-07
Miguel 3.1 - Three in one [Python]

miguelHx:
Implementation of triple stack is in StackTrio class, corresponding test cases at the bottom


Python/chapter03/p01_three_in_one/miguelHx.py
```diff
@@ -0,0 +1,420 @@
+"""
+Python version 3.7.0
```
lhchavez, line 2: FYI: Python multi-line docstrings don't  have a newline after the opening `"""` per https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings (but do have a newline before the closing `"""`, so L5 is fine).

Python/chapter03/p01_three_in_one/miguelHx.py
```diff
@@ -0,0 +1,420 @@
...
+3.1 - Three in one
+Describe how you could use a single array to implement three stacks
+"""
+
+import copy
+import unittest
+from typing import List
+
+
+class StackNode(object):
```
lhchavez, line 12: FYI: the `(object)` pattern was somewhat required in Python2. Python3 has fixed all the footguns associated with leaving it off and some folks prefer it that way!

also, if this is a data class, might be better to make it a real [`dataclasses.dataclass`](https://docs.python.org/3/library/dataclasses.html):

```python
from dataclasses import dataclass
from typing import Generic, TypeVar

T = TypeVar('T')

@dataclass
class StackNode(Generic[T]):
    data: T
    next: 'StackNode[T]'
```

dataclasses force you to use types (which is a good thing!), which help in documentation. i didn't need to add a comment to what `next` is, because the type already conveys that information, for both humans and machines.

Python/chapter03/p01_three_in_one/miguelHx.py
```diff
@@ -0,0 +1,420 @@
...
+class MyStack(object):
+    """
+    Stack data structure implementation.
+    Uses LIFO (last-in first-out) ordering.
+    The most recent item added to the stack is
+    the first removed.  Traversal is top to bottom.
+    """
+
+    def __init__(self):
+        self.top = None # top is a pointer to StackNode object
```
lhchavez, line 27:
```suggestion
        self.top: Optional[StackNode[T]] = None
```

it might be a good idea to use [type annotations](https://docs.python.org/3/library/typing.html) to convey this information, and might as well add them to all the public interface. it helps both humans (to quickly understand what's going on) and machines (with [mypy](http://mypy-lang.org/), code can be super quickly validated for type-correctness, which catches some families of bugs at "compile time")! static analysis FTW!

Python/chapter03/p01_three_in_one/miguelHx.py
```diff
@@ -0,0 +1,420 @@
...
+        self.size = 0
+        return
+    
+    def pop(self) -> int:
```
lhchavez, line 31:
```suggestion
    def pop(self) -> T:
```

the class can be declared as `class MyStack(Generic[T])` to let any type (not just integers) be stored.

Python/chapter03/p01_three_in_one/miguelHx.py
```diff
@@ -0,0 +1,420 @@
...
+            IndexError: [description]
+
+        Returns:
+            int: the value at the top of the stack
+        """
+        if self.top is None:
+            raise IndexError('Stack is Empty')
+        return self.top.data
+    
+    def is_empty(self) -> bool:
```
lhchavez, line 73: maybe this can use [`__bool__`](https://docs.python.org/3/reference/datamodel.html#object.__bool__) so that instead of `if !my_list.is_empty():` one can use `if my_list:`.

Python/chapter03/p01_three_in_one/miguelHx.py
```diff
@@ -0,0 +1,420 @@
...
+        """
+        Returns True if and only if the
+        stack is empty, otherwise False.
+
+        Returns:
+            bool: True if empty, False otherwise
+        """
+        return self.size == 0
+    
+    def as_list(self) -> List[int]:
```
lhchavez, line 83: optional: if this class implements [`__iter__`](https://docs.python.org/3/library/stdtypes.html#typeiter), this can be expressed as `list(my_stack)`! it also has the benefit of being able to be converted to any other container type that supports an iterable as argument (e.g. `set(my_stack)`).

Python/chapter03/p01_three_in_one/miguelHx.py
```diff
@@ -0,0 +1,420 @@
...
+        self.num_stacks = 3
+        self.stack_capacity = stack_capacity
+        self.stack_info = {
+            1: {
+                'start': 0,
+                'end': stack_capacity - 1,
+                'size': 0,
+                'top_index': 0,
+                'top_index_next': -1, # index after top, initially starts at a sentinel value because the stack is empty, so there is no "next"
+            },
```
lhchavez, line 125: it _may_ be better to make this data structure a class/dataclass: that way the entries can be referred to by `.start` instead of `['start']`.

(also, mypy can help catch small bugs!)

Python/chapter03/p01_three_in_one/miguelHx.py
```diff
@@ -0,0 +1,420 @@
...
+        while n.next is not None:
+            values.append(str(n.data))
+            n = n.next
+        values.append(str(n.data))
+        return '->'.join(values)
+
+
+class StackTrio(object):
+
+    def __init__(self, stack_capacity = 7):
```
lhchavez, line 115: this is a good first approach! having a fixed-sized array is a good way of doing this.

optional stretch goal: there are several ways to support this without specifying an upfront capacity.

Python/chapter03/p01_three_in_one/miguelHx.py
```diff
@@ -0,0 +1,420 @@
...
+                'end': stack_capacity * 3 - 1,
+                'size': 0,
+                'top_index': stack_capacity * 2,
+                'top_index_next': -1
+            }
+        }
+        self.values = [0] * (stack_capacity * self.num_stacks)
+
+    def is_empty(self, stack_id):
+        return self.stack_info[stack_id]['size'] <= 0
```
lhchavez, line 144: what happens if a caller provides a `stack_id` that's not in the range (1...3)?

Python/chapter03/p01_three_in_one/miguelHx.py
```diff
@@ -0,0 +1,420 @@
...
+from typing import List
+
+
+class StackNode(object):
+
+    def __init__(self, data):
+        self.data = data
+        self.next = None # next is a pointer to a StackNode object
+
+class MyStack(object):
```
lhchavez, line 18: seems like this class is not being used in the solution of the problem (StackTrio)

Python/chapter03/p01_three_in_one/miguelHx.py
```diff
@@ -0,0 +1,420 @@
...
+        # decremented by one. (Or, each get decremented by 1)
+
+        # get stack top index, then set value to 0 as a way of clearing
+        original_stack_top_index = self.stack_info[stack_id]['top_index']
+        val_before_pop = self.peek(stack_id)
+
+        # special case, when there is only one element in this stack,
+        # we will clear top index, and reset next to sentinel value (-1)
+
+        if self.get_size(stack_id) == 1:
```
lhchavez, line 200: is there a way in which this special case can be avoided?

Python/chapter03/p01_three_in_one/miguelHx.py
miguelHx: Updated

Python/chapter03/p01_three_in_one/miguelHx.py
miguelHx: Updated, this is neat 👍

Python/chapter03/p01_three_in_one/miguelHx.py
miguelHx: Fixed in b84c365dc814a76fb065081d91b604bab2eb32a6

Python/chapter03/p01_three_in_one/miguelHx.py
miguelHx: Updated f0767768f1ad7ccbcabf2aba6e260b358ae44d1e

Python/chapter03/p01_three_in_one/miguelHx.py
miguelHx: Implemented here:  63133fd55f83cf57a4bfe55c711e4bbd02455f1b

Python/chapter03/p01_three_in_one/miguelHx.py
miguelHx: Thank you!  may revisit the optional stretch goal later

Python/chapter03/p01_three_in_one/miguelHx.py
miguelHx: Good catch 👍  fixed in cb1fe7577bd3dbfc8ea3f6ec4915e3674a3f11c9

Python/chapter03/p01_three_in_one/miguelHx.py
miguelHx: It is not.  But StackTrio uses StackNode, which MyStack also uses.  I figured that if changes would potentially need to be made to StackNode, then I'll make updates to both MyStack and StackTrio in the same file.

The reason why I did not leave MyStack in a separate file and import it is because I ran into problems with relative imports.

lhchavez: oh yes, relative imports don't work _at all_ outside of packages :(

most codebases tend to prefer removing code that is not currently used, and deferring its introduction to the moment where they start being used. in this case, the `MyStack` class could safely be removed from this solution and appear only in 3.2.

`StackNode` can be kept.

Python/chapter03/p01_three_in_one/miguelHx.py
```diff
@@ -0,0 +1,445 @@
...
+            self.current_node = self.top
+            raise StopIteration
+        n: T = self.current_node.data
+        self.current_node = self.current_node.next
+        return n
+    
+    def __bool__(self) -> bool:
+        """
+        Returns True if and only if the
+        stack is empty, otherwise False.
```
lhchavez, line 105: idiomatically in Python, `True` is returned when the container is _not_ empty. From https://docs.python.org/3/reference/datamodel.html#object.__bool__ : 

> Called to implement truth value testing and the built-in operation bool(); should return False or True. When this method is not defined, __len__() is called, if it is defined, and the object is considered true if its result is nonzero. If a class defines neither __len__() nor __bool__(), all its instances are considered true.

Python/chapter03/p01_three_in_one/miguelHx.py
miguelHx: Updated here: 098292e6fb4390059ef567a031e45053bee74735

Python/chapter03/p01_three_in_one/miguelHx.py
miguelHx: This code was brought in based on the idea of "resetting" stack state which doesn't go back into its original "new" state.  That is, setting the next index as the sentinel value (-1)....Even though it wouldn't contribute to the correctness of the other functions. 

For an empty stack, it doesn't matter what the starting value of `top_index_next` is because when we push an element onto an empty stack, the `top_index_next` will be created based off the value of `top_index`.  So, all that matters is that `top_index` is reset correctly, for both `top_index_next` and `top_index` to work.

This means that the code I added was redundant.  My thinking was too narrow while writing that.
The solution will be to remove code related to resetting `top_index_next`.

miguelHx: FIxed here:  7a4d520912577a65200baa9f5670a6c3dc698edd

Python/chapter03/p01_three_in_one/miguelHx.py
miguelHx: Good catch.  I somehow forgot to add proper test cases for __bool__.

Resolved in e07edab577b0c84bee942fd9e7d8975adb9e9376

Python/chapter03/p01_three_in_one/miguelHx.py
miguelHx: Resolved with a dataclass here:  a33465cdd679555af66311c8af8244d3c0bbbb41

Python/chapter03/p01_three_in_one/miguelHx.py
```diff
@@ -17,137 +17,49 @@ class StackNode(Generic[T]):
...
     def __init__(self, stack_capacity = 7):
         self.num_stacks = 3
         self.stack_capacity = stack_capacity
+        first_stack_info = StackInfo(
+            1, # id
+            0, # start index
+            stack_capacity - 1, # end index
+            0, # size
+            0, # top index
+            -1 # top index next
```
lhchavez, line 41:
```suggestion
            id: 1,
            start_index: 0,
            end_index: stack_capacity - 1,
            size: 0,
            top_index: 0,
            top_next_index: -1
```

this way the comments are not needed anymore, _and_ the parameters can be reordered safely.

Python/chapter03/p01_three_in_one/my_stack.py
```diff
@@ -0,0 +1,183 @@
...
+
+T = TypeVar('T')
+
+@dataclass
+class StackNode(Generic[T]):
+    data: T
+    next: 'Optional[StackNode[T]]'
+
+
+class MyStack:
```
lhchavez, line 21:
```suggestion
class MyStack(Generic[T]):
```

otherwise mypy will complain that this class is not generic when attempting to use `T` in L30

Python/chapter03/p01_three_in_one/miguelHx.py
miguelHx: This led to a syntax error.  Seems like Python 3 doesn't support that syntax when initializing an instance of a dataclass.

Maybe the best way for now is to use kwargs?  Updated to use kwargs here:  6f0999438665d195fe19ccc3ae4adb00f80a2388

Python/chapter03/p01_three_in_one/my_stack.py
miguelHx: Line 30 uses T in `StackNode[T]` but no error messages from mypy 🤔 .  With or without `Generic[T]`.  After some investigation, I cannot seem to find why this is the case.  I'm using mypy version `0.910`.

Either way, going to commit your suggested change 👍

Python/chapter03/p01_three_in_one/miguelHx.py
lhchavez: oh derp! i forgot it's with `=`, not `:`! this is a function call, not a dictionary u_u

(note to self: try out suggestions locally)

miguelHx: No worries, is this PR ready to be merged?

lhchavez (approved):

PR #83, opened 2021-08-10
Miguel 3.2 - Stack Min [Python]

miguelHx:


Python/chapter03/p02_stack_min/miguelHx.py
```diff
@@ -0,0 +1,222 @@
...
+        with self.assertRaises(IndexError):
+            s.pop()
+        s.push(1)
+        s.push(2)
+        s.push(3)
+        # size is 3
+        self.assertEqual(s.as_list(), [3, 2, 1])
+        val = s.pop()
+        self.assertEqual(val, 3)
+        self.assertEqual(s.size, 2) # size should now be 2
```
lhchavez, line 186: optional: it might be better to make `MyStack` implement [`__len__`](https://docs.python.org/3/reference/datamodel.html#object.__len__) and rename `size` to `_size`: that way callers of the class can't mess around with internals.

Python/chapter03/p02_stack_min/miguelHx.py
```diff
@@ -0,0 +1,222 @@
...
+class MyStack(object):
+    """
+    Stack data structure implementation.
+    Uses LIFO (last-in first-out) ordering.
+    The most recent item added to the stack is
+    the first removed.  Traversal is top to bottom.
+    """
+
+    def __init__(self):
+        self.top = None # top is a pointer to StackNode object
```
lhchavez, line 31: same suggestions w.r.t. typing as with https://github.com/techqueria/data-structures-and-algorithms/pull/82

Python/chapter03/p02_stack_min/miguelHx.py
```diff
@@ -0,0 +1,222 @@
...
+        self.size += 1
+
+    def min(self) -> int:
+        """
+        Returns the element with the lowest value in the stack
+
+        Returns:
+            int: min value in stack
+        """
+        return self.top.running_min
```
lhchavez, line 81: what happens if `self.top` is `None`?

Python/chapter03/p02_stack_min/miguelHx.py
miguelHx: Updated here 64d2effe52c84e024dc164a6e2b2e9356d485e21

Python/chapter03/p02_stack_min/miguelHx.py
miguelHx: Will get an exception for non existing property on None type.

Resolved in ae680053e33e944620aaef98b9c1d6562f49c6c0

Python/chapter03/p02_stack_min/miguelHx.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        t = StackNode(item)
+        t.next = self.top
+        if self.top is None:
+            t.running_min = item
+            self.top = t
+            self.size += 1
+            # print("Stack Node data = {}, local min = {}".format(item, t.running_min))
+            return
+        if item < self.top.data:
```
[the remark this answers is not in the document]
miguelHx: Resolved in ae680053e33e944620aaef98b9c1d6562f49c6c0

Python/chapter03/p02_stack_min/miguelHx.py
miguelHx: Good idea, resolved in 745f6a4e1b08a36c0808435afd3c4bec186e8b5e

miguelHx:
Test cases are passing now but I am getting an error when running `mypy`:

```
miguelHx.py:67: error: Unsupported left operand type for < ("T")
Found 1 error in 1 file (checked 1 source file)
```

Made an attempt to fix in d87e76cf6be77ef715711b46ddb5d6e1f56a06fe

Python/chapter03/p02_stack_min/miguelHx.py
```diff
@@ -0,0 +1,248 @@
...
+from typing import Generic, TypeVar
+from typing import List, Optional, Iterator
+
+T = TypeVar('T')
+C = TypeVar('C', bound='Comparable')
+
+class Comparable(Generic[T, C]):
+    @abstractmethod
+    def __lt__(self, other: T):
+        pass
```
lhchavez, line 24:
```suggestion
from typing import Protocol

T = TypeVar('T', bound='Comparable')

class Comparable(Protocol):
    @abstractmethod
    def __lt__(self, other: T) -> bool:
        pass
```

from https://github.com/python/typing/issues/59#issuecomment-353878355

`C` isn't used as a type anywhere, so it can go.

Python/chapter03/p02_stack_min/miguelHx.py
```diff
@@ -0,0 +1,248 @@
...
+
+@dataclass
+class StackNode(Generic[T, C]):
+    data: T
+    next: 'Optional[StackNode[T, C]]'
+    running_min: T
```
lhchavez, line 30:
```suggestion
class StackNode(Generic[T]):
    data: T
    next: 'Optional[StackNode[T]]'
    running_min: T
```

Python/chapter03/p02_stack_min/miguelHx.py
```diff
@@ -0,0 +1,248 @@
...
+    Uses LIFO (last-in first-out) ordering.
+    The most recent item added to the stack is
+    the first removed.  Traversal is top to bottom.
+    """
+
+    def __init__(self):
+        self.top: Optional[StackNode[T]] = None # top is a pointer to StackNode object
+        self._size: int = 0
+        self.index: int = -1 # for iterator use
+        self.current_node: Optional[StackNode[T]] = self.top # also for iterator use
```
lhchavez, line 43: optional: the most idiomatic way to express this is to have an additional class (maybe even embedded in this class as an inner class) that holds these members instead of the main class. this allows for multiple concurrent iterators:

https://anandology.com/python-practice-book/iterators.html#the-iteration-protocol

Python/chapter03/p02_stack_min/miguelHx.py
```diff
@@ -0,0 +1,248 @@
...
+        return
```
lhchavez, line 44: optional:

```suggestion
```

this is not needed.

Python/chapter03/p02_stack_min/miguelHx.py
```diff
@@ -0,0 +1,248 @@
...
+    def __lt__(self, other: T):
+        pass
+
+@dataclass
+class StackNode(Generic[T, C]):
+    data: T
+    next: 'Optional[StackNode[T, C]]'
+    running_min: T
+
+class MyStack(object):
```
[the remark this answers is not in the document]
miguelHx: Same response as in #82:

> Line 30 uses T in `StackNode[T]` but no error messages from mypy 🤔 .  With or without `Generic[T]`.  After some investigation, I cannot seem to find why this is the case.  I'm using mypy version `0.910`.
> Either way, going to commit your suggested change 👍

Python/chapter03/p02_stack_min/miguelHx.py
miguelHx: Committed suggested change here: 3c8ceb5a7249732782b5828321325233cb062008
Removed remaining C usage here:  6d9c43e99e208cde9e734e24591a721c5b2d5c5a

Python/chapter03/p02_stack_min/miguelHx.py
miguelHx: Updated here 6d9c43e99e208cde9e734e24591a721c5b2d5c5a in response to another comment

Python/chapter03/p02_stack_min/miguelHx.py
miguelHx: Implemented here:  472f8f499950d068d2869b5ea4574cddeb7eb77d

Python/chapter03/p02_stack_min/miguelHx.py
miguelHx: Removed in 472f8f499950d068d2869b5ea4574cddeb7eb77d

lhchavez (approved):

PR #85, opened 2021-08-19
Miguel 3.4 - Queue via Stacks [Python] 

miguelHx:


Python/chapter03/p04_queue_via_stacks/miguelHx.py
```diff
@@ -0,0 +1,290 @@
...
+
+    def peek(self) -> T:
+        if self._size == 0:
+            raise Exception('No values in stack to remove')
+        return self._extract_first_in('peek')
+
+    def is_empty(self) -> bool:
+        return self._size == 0
+    
+    def _extract_first_in(self, operation: str) -> T:
```
lhchavez, line 159: in cases like this, it's preferable to use [`enum.Enum`](https://docs.python.org/3/library/enum.html#creating-an-enum) to make the options type-safe.

it may be even more preferable to split this in two functions (one to pop and another one to peek) that call other helper functions to do 1 and 3 (and in fact can be done with a single helper function where the stacks are passed as parameters). that way the enum is not even needed.

Python/chapter03/p04_queue_via_stacks/miguelHx.py
```diff
@@ -0,0 +1,290 @@
...
+        return '->'.join(values)
+
+
+@dataclass
+class QueueNode(Generic[T]):
+    data: T
+    next: 'Optional[QueueNode[T]]'
+
+
+class MyQueue:
```
lhchavez, line 133:
```suggestion
class MyQueue(Generic[T]):
```

this way it's legal to refer to `T` inside this class.

Python/chapter03/p04_queue_via_stacks/miguelHx.py
```diff
@@ -0,0 +1,290 @@
...
+    def __init__(self):
+        self._size = 0
+        self.stack_one = MyStack()  # will be used as main data container
```
lhchavez, line 136:
```suggestion
        self.stack_one = MyStack[T]()  # will be used as main data container
```

Python/chapter03/p04_queue_via_stacks/miguelHx.py
```diff
@@ -0,0 +1,290 @@
...
+        self.stack_two = MyStack()  # will be used to extract the least recent insertion
```
lhchavez, line 137:
```suggestion
        self.stack_two = MyStack[T]()  # will be used to extract the least recent insertion
```

Python/chapter03/p04_queue_via_stacks/miguelHx.py
lhchavez: optional: by calling this `self.data_stack`, the comment about this being a data container is no longer needed.

Python/chapter03/p04_queue_via_stacks/miguelHx.py
lhchavez: optional: this is only needed when the peek/pop operations are done as a helper variable. it can be removed from the list of members and created as a temporary variable inside those methods.

Python/chapter03/p04_queue_via_stacks/miguelHx.py
miguelHx: Split into two functions here:  4e38eabef4b023618802ec82779d57318bf16a47

Python/chapter03/p04_queue_via_stacks/miguelHx.py
miguelHx: Updated here 4e38eabef4b023618802ec82779d57318bf16a47

Python/chapter03/p04_queue_via_stacks/miguelHx.py
miguelHx: Updated here:  d3eca76f03c5cfc69c393ef8d44ddfa654c77208

Python/chapter03/p04_queue_via_stacks/miguelHx.py
miguelHx: Updated here:  d3eca76f03c5cfc69c393ef8d44ddfa654c77208

lhchavez (approved):

PR #87, opened 2021-08-23
Miguel 3.6 - Animal Shelter [Python]

miguelHx:


Python/chapter03/p06_animal_shelter/miguelHx.py
```diff
@@ -0,0 +1,222 @@
...
+class AnimalTypeEnum(Enum):
+    DOG = 'dog'
+    CAT = 'cat'
+
+
+@dataclass
+class Animal:
+    name: str
+    species: AnimalTypeEnum
+    timestamp: Optional[int] = None
```
lhchavez, line 27: from the typing perspective, the timestamp is not an inherent property of the animal, and there is nothing preventing a caller from fiddling with the timestamp from the outside, confusing the implementation. furthermore, it is not possible to express that an animal that currently exists in the shelter is guaranteed to have a timestamp purely from the type system.

there _is_ a way to solve both of these problems! by creating a wrapper class

```python
@dataclass
class QueuedAnimal:
    animal: Animal
    timestamp: int
```

and wrapping `Animal`s into `QueuedAnimal`s upon enqueueing.

Python/chapter03/p06_animal_shelter/miguelHx.py
```diff
@@ -0,0 +1,222 @@
...
+        self._global_timestamp = 0
+        self._size = 0
+    
+    def _enqueue(self, animal: Animal) -> None:
+        # list will grow right [<animal_1>, <animal_2>, ..., <animal_n>]
+        animal.timestamp = self._global_timestamp
+        self._global_timestamp += 1
+        self._species_map[animal.species].append(animal)
+
+    def enqueue(self, *animals: Animal, key: str = '') -> None:
```
lhchavez, line 53: what's the key for?

Python/chapter03/p06_animal_shelter/miguelHx.py
```diff
@@ -0,0 +1,222 @@
...
+        self.assertEqual(len(a.dogs), 1)
+        self.assertEqual(len(a.cats), 0)
+        self.assertEqual(d2, d)
+        d = a.dequeueAny()
+        self.assertEqual(len(a), 0)
+        self.assertEqual(len(a.dogs), 0)
+        self.assertEqual(len(a.cats), 0)
+        self.assertEqual(d3, d)
+        self.assertRaises(IndexError, lambda: a.dequeueAny())
+        # scenario 2: Only cats
```
lhchavez, line 176: given that every scenario is (logically) independent of the previous ones, it may be better to split this test in four for readability. also, if the test fails, a maintainer would need to look at the test code and read the position of the failure, plus fix the failure to discover whether the following behaviors exercised in the test still work.

It is a best practice to make tests as small and isolated as possible. Per https://dzone.com/articles/unit-testing-best-practices-how-to-get-the-most-ou

> #### Unit Tests Should Verify a Single-Use Case
> Good tests validate one thing and one thing only, which means that typically, they validate a single use-case. Tests that follow this best practice are simpler and more understandable, and that is good for maintainability and debugging. Tests that validate more than one thing can easily become complex and time-consuming to maintain. Don't let this happen.

Python/chapter03/p06_animal_shelter/miguelHx.py
miguelHx: Thank you !  Updated here:  ffb663aad36f9395ae08912bdc2513bfb2832d11

How does one get really good at designing classes?

Python/chapter03/p06_animal_shelter/miguelHx.py
miguelHx: That was mean't to be taken out, left it there by accident.  It was based on my pre-initial implementation idea where key would be 'dog' or 'cat' and that would be used with some conditional logic.  Something like:
python
```
if key == 'dog':
    self.dogs.append(animal)
if key == 'cat':
    self.cats.append(animal)
```
Soon after I decided it would be better to reference the species information from the animal object
rather than using a redundant 'key' argument.

I was meaning to get rid of it but somehow let it slip.

Removed in 93df0f4509874ba862afd9a6abd1e37834f7c009

Python/chapter03/p06_animal_shelter/miguelHx.py
miguelHx: Great feedback, thank you!  Updated here:  a88c163df248d3ac2735534ba40ef02e3af68626

Python/chapter03/p06_animal_shelter/miguelHx.py
lhchavez: There are two ideas that can give you very good bang for the buck:

* Make interfaces hard to misuse: try to constrain the options available to callers to avoid them (accidentally) falling into mistakes. This can be achieved by providing higher-level methods (instead of requiring the caller to chain several lower-level methods in sequence), removing unnecessary fields from the public interface, creating factory functions that return fully-initialized objects (instead of relying on docstrings/examples that tell the caller how to initialize it), bundling sets of required parameters into classes or structures to make it obvious that they _need_ to be all defined together, etc. There's a good little article here: https://testing.googleblog.com/2018/07/code-health-make-interfaces-hard-to.html
* There's an idiom in functional programming called "make illegal states unrepresentable", which means that folks should design their data structures in a way that invalid combinations of fields in a class cannot even be expressed. There's a good article about it here, taking it further and applying this idea in TypeScript through Domain-Driven Design: https://khalilstemmler.com/articles/typescript-domain-driven-design/make-illegal-states-unrepresentable/

lhchavez: oh right, i forgot: there's also discouraging the use of `Optional`: inside a class, every `Optional` makes the number of representable states to double! odds are, some of those 2^n combinations aren't valid. by moving the fields marked `Optional` to another class (like we did in this thread), or bundling them all into a class to signify that they must all be provided or are all optional _together_ (say, there's a class that allows a different functionality for anonymous users and logged in users that have completed their profile, the set of (session_id, user_id, profile) can be bundled into a class that _itself_ can be marked as `Optional` altogether).

lhchavez (approved):

PR #88, opened 2021-08-29
Miguel 4.1 - Route Between Nodes [Python]

miguelHx:
Critical feedback is welcome for all of my PRs


Python/chapter04/p01_route_between_nodes/miguelHx.py
```diff
@@ -0,0 +1,202 @@
...
+class Graph:
+    nodes: 'List[Node]'
+
+    def print_graph(self):
+        for node in self.nodes:
+            node.print_children()
+
+    def reset_visited(self):
+        for node in self.nodes:
+            node.visited = False
```
lhchavez, line 23: it's probably better to store the visitedness of the nodes outside them. that way future algorithms can reuse this data structure and use it _slightly_ differently. for instance, some algorithms require coloring nodes instead of a boolean flag (colors could be seen as a generalization of visitedness).

this could be achieved by having `visited: Set[int] = set()` and having that visited set store the `id`s of the nodes.

Python/chapter04/p01_route_between_nodes/miguelHx.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def add_child(self, *nodes: 'Node'):
+        for node in nodes:
+            self.children.append(node)
+
+    def print_children(self):
+        message = f'Adjacency list for node ({self.id}): '
+        for child in self.children:
+            message += '{}, '.format(child.id)
+        print(message)
```
lhchavez, line 40: for performance reasons, it's better to avoid concatenating to a string in a loop:

```suggestion
        logging.debug('Adjacency list for node %s: %s', self.id, ', '.join(str(child.id) for child in self.children))
```

also, the [`logging`](https://docs.python.org/3/library/logging.html) module is preferred for this, because it's configurable.

Python/chapter04/lib/graph_search.py
```diff
@@ -0,0 +1,143 @@
+import unittest
```
lhchavez, line 1: is this file used by the solution? if not, it can be removed

Python/chapter04/p01_route_between_nodes/miguelHx.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def __str__(self):
+        return f'Node ({self.id}), visited: {self.visited}'
+
+
+def dfs_search(root: Node) -> List[int]:
```
lhchavez, line 46: is this part of the solution? if not, it can be removed.

Python/chapter04/p01_route_between_nodes/miguelHx.py
```diff
@@ -0,0 +1,202 @@
...
+
+    Returns:
+        List[int]: List[int]: list of node IDs (i.e. [0, 1, 4])
+    """
+    visited_nodes: List[Node] = []
+    visited_nodes.append(root)
+    queue: Deque[Node] = deque()
+    root.visited = True
+    queue.append(root)
+    while len(queue) >= 1:
```
lhchavez, line 87:
```suggestion
    while queue:
```

`deque` implements the `__bool__` magic method.

Python/chapter04/p01_route_between_nodes/miguelHx.py
```diff
@@ -0,0 +1,202 @@
...
+    nodes.
+
+    Args:
+        root (Node): starting node
+
+    Returns:
+        List[int]: List[int]: list of node IDs (i.e. [0, 1, 4])
+    """
+    visited_nodes: List[Node] = []
+    visited_nodes.append(root)
```
lhchavez, line 83:
```suggestion
    visited_list: List[int] = [root.id]
    visited: Set[int] = set([root.id])
```

that way the `Node.visited` boolean is not needed anymore, and L99 can return `visited_list` as-is.

Python/chapter04/p01_route_between_nodes/miguelHx.py
```diff
@@ -0,0 +1,202 @@
...
+    queue: Deque[Node] = deque()
+    root.visited = True
+    queue.append(root)
+    while len(queue) >= 1:
+        node = queue.popleft()
+        node.visited = True
```
lhchavez, line 89: given that the visited flag is set _before_ adding any element to the queue, this can go.

```suggestion
```

Python/chapter04/p01_route_between_nodes/miguelHx.py
```diff
@@ -0,0 +1,202 @@
...
+        root (Node): starting node
+
+    Returns:
+        List[int]: List[int]: list of node IDs (i.e. [0, 1, 4])
+    """
+    visited_nodes: List[Node] = []
+    visited_nodes.append(root)
+    queue: Deque[Node] = deque()
+    root.visited = True
+    queue.append(root)
```
lhchavez, line 86:
```suggestion
    queue: Deque[Node] = deque([root])
```

the [`deque` constructor](https://docs.python.org/3/library/collections.html#collections.deque) can take an iterable.

Python/chapter04/p01_route_between_nodes/miguelHx.py
```diff
@@ -0,0 +1,202 @@
...
+
+    Args:
+        src (Node): from node
+        dest (Node): destination node
+
+    Returns:
+        bool: whether a path between src and dest exists
+    """
+    ids_visited: List[int] = bfs_search(src)
+    return True if dest.id in ids_visited else False
```
lhchavez, line 126:
```suggestion
    return dest.id in ids_visited
```

there's no need for the "ternary"!

Python/chapter04/p01_route_between_nodes/miguelHx.py
```diff
@@ -0,0 +1,202 @@
...
+        in the graph.
+
+    Args:
+        src (Node): from node
+        dest (Node): destination node
+
+    Returns:
+        bool: whether a path between src and dest exists
+    """
+    ids_visited: List[int] = bfs_search(src)
```
lhchavez, line 125: small optimization: in very large graphs, the search can be terminated earlier the moment the destination node is reached. same worst-case complexity, but it's a tiny bit faster.

Python/chapter04/p01_route_between_nodes/miguelHx.py
miguelHx: 265cf29b04bfc1e75ac795c3422ee831d68fe02c
Node.visited no longer needed

Python/chapter04/p01_route_between_nodes/miguelHx.py
miguelHx: 39c084ce4086aee21900712cd5fd9e7574e0ddd0

Python/chapter04/p01_route_between_nodes/miguelHx.py
miguelHx: Right, I remember that conversation we had before about concatenating to a string in a loop.  .join is preferred
Committed your suggestion here:  2ef8112845a08a2c0c804b221b286f423c786197

Python/chapter04/lib/graph_search.py
miguelHx: Not used directly by the solution.  Going to remove.  Moved this code to a separate PR to make it easier for me to reference commonly used code

miguelHx: File removed here:  59bceeefc2449efb37ef2aa5a948e234e2b8ec66

Python/chapter04/p01_route_between_nodes/miguelHx.py
miguelHx: Removed usage of .visited here:  75c59a4fbe9b2495af76d09a46092afd5e1da4f8

Python/chapter04/p01_route_between_nodes/miguelHx.py
miguelHx: Good call, updated here:  e5f27c89ce8531ba096309714f2a91dff75de95b

lhchavez (approved):

